PR #1, opened 2020-11-08
First draft of a module "summarise"

nevrome:
Please review strictly - I'm sure it can be improved a lot.

Regarding the validation feature with useful, verbose user feedback that shows multiple exceptions: Cassava does not seem to have a built-in solution for that, but maybe one could still achieve this with line-wise reading. So instead of parsing the whole `readFile` bytestring, one could split it up into lines and then continue gracefully if reading of a line fails. Could this work?


stschiff:
Sieht schonmal echt super aus! Ich melde mich mit Details.

src/Poseidon/Package.hs
```diff
@@ -280,3 +330,62 @@ compFunc1 (EigenstratSnpEntry c1 p1 _ _ _ _, _) (EigenstratSnpEntry c2 p2 _ _ _
 compFunc2 :: (EigenstratSnpEntry, GenoLine) -> [(EigenstratSnpEntry, GenoLine)] -> Ordering
 compFunc2 (EigenstratSnpEntry c1 p1 _ _ _ _, _) ((EigenstratSnpEntry c2 p2 _ _ _ _, _):_) = compare (c1, p1) (c2, p2)
 compFunc2 _                                     []                                        = error "compFunc2 - should never happen"
+
+-- Janno file loading
+
+decodingOptions :: Csv.DecodeOptions
+decodingOptions = Csv.defaultDecodeOptions { 
+    Csv.decDelimiter = fromIntegral (ord '\t')
```
stschiff, line 338: This is bizarre... `fromIntegral . ord` must be basically `id`. I'll have a look what type `Csv.decDelimiter` needs to be.

src/Poseidon/Package.hs
```diff
@@ -280,3 +330,62 @@ compFunc1 (EigenstratSnpEntry c1 p1 _ _ _ _, _) (EigenstratSnpEntry c2 p2 _ _ _
...
+}
+
+instance Csv.FromRecord PoseidonSample
+
+-- | A helper function for parsing double lists
+bytestringToDouble :: [Bch.ByteString] -> [Double]
+bytestringToDouble [] = []
+bytestringToDouble (x:xs) = (read (Bch.unpack x) :: Double) : bytestringToDouble xs
```
stschiff, line 346: OK, using `read` is generally considered unsafe because this will fail ungracefully if the unpacked string can't be read as an integer. Have you checked what happens in such a case?

src/Poseidon/Package.hs
```diff
@@ -280,3 +330,62 @@ compFunc1 (EigenstratSnpEntry c1 p1 _ _ _ _, _) (EigenstratSnpEntry c2 p2 _ _ _
...
+
+instance Csv.FromField [Double] where
+    parseField = fmap bytestringToDouble . fmap (\x -> Bch.splitWith (==';') x) . Csv.parseField
+
+-- | A helper function for parsing integer lists
+bytestringToInteger :: [Bch.ByteString] -> [Integer]
+bytestringToInteger [] = []
+bytestringToInteger (x:xs) = (read (Bch.unpack x) :: Integer) : bytestringToInteger xs
```
stschiff, line 354: Same here... maybe the error gets somehow propagated to the general CSV reading routine and things are fine. In any case this wouldn't prevent us from merging this PR and fixing later, but it's something to keep an eye on. We anyway need some automatic tests for Janno parsing.

stschiff:
Do you have a mechanism for checking that fields where only specific strings are allowed ("choice_options") are correct?

src/Poseidon/Package.hs
```diff
@@ -280,3 +330,62 @@ compFunc1 (EigenstratSnpEntry c1 p1 _ _ _ _, _) (EigenstratSnpEntry c2 p2 _ _ _
...
+bytestringToString :: [Bch.ByteString] -> [String]
+bytestringToString [] = []
+bytestringToString (x:xs) = (Bch.unpack x) : bytestringToString xs
+
+instance Csv.FromField [String] where
+    parseField = fmap bytestringToString . fmap (\x -> Bch.splitWith (==';') x) . Csv.parseField
+
+-- | A utility function to load multiple janno files
+loadJannoFiles :: [FilePath] -> IO [[PoseidonSample]]
+loadJannoFiles jannoPaths = do
```
stschiff, line 369: (Not so important, but...) this can be written point-free as
```
loadJannoFiles = mapM loadJannoFile
```

nevrome:
Not yet. Should this be solved with a custom data type for the respective variables?

stschiff (approved):
OK, very nice! As you saw I have put some minor comments, but I guess we can address most of them later as well. We also need automatic testing of janno-file parsing, which is some work, as we need to provide some fake Janno Files for the fake testModules in the test Directory. Perhaps we can just copy-paste some rows from real janno files and write some tests.

stschiff:
OK, approved. I'm happy to merge this in now, unless you'd like to push a quick commit regarding my comments in the code (none of them are critical for now, except for the unsafe `read` thing, but that can still be ironed out once we have tests).

src/Poseidon/Package.hs
stschiff: OK, found it. The docs even recommend this, because the type of `Csv.decDelimiter` is a `Word8`. Fine.

stschiff:
> Not yet. Should this be solved with a custom data type for the respective variables?

Yes. Exactly, that would be ideal. But I'm happy to a address that later if it's unclear how to build such custom field parsers in cassava.

nevrome:
Alright - thank you - I will see what I can do!

nevrome:
Is 8ed1854 already what you had in mind for these choice fields?

nevrome:
The solution with specific data types seems to work for choice fields. Should we also use it for fields with range limits? 

An entry in the field `Latitude` should be between -90 and +90. I guess one could define a datatype with a restrictive constructor to prevent wrong input in this field from being loaded.

src/Poseidon/Package.hs
```diff
@@ -75,10 +84,111 @@ data GenotypeDataSpec = GenotypeDataSpec
...
+
+-- |A datatype to represent Library_Built in a janno file
+data JannoLibraryBuilt = 
+      DS
+    | SS
+    | Other
+    deriving (Eq, Show, Ord)
+
+-- | A datatype for Latitudes
+data Latitude =
```
stschiff, line 136: In so einem Fall lohnt sich übrigens `newtype` statt `data`. Das sorgt dafür dass der Compiler das intern trotzdem als Double darstellt, nur eben type safe. Ist also performanter. `newtype` lohnt sich immer dann wenn man einen abstrakten Datentypen über einem built-in Datentypen baut, mit einem einfachen Constructor. Nur so als Tipp...

PR #28, opened 2021-01-08
packageVersion

### LINKED ISSUE #23: packageVersion
Poseidon v.2.0.1 states that each POSEIDON.yml file should have a `packageVersion` field that should be incremented with changes to the respective package. poseidon-hs ignores this field so far.

nevrome:
I added the packageVersion field for the POSEIDON.yml and made the duplicate search depend on it (instead of on lastModified). 

If merged this closes #23.


stschiff (approved):
Good to go.

Issue #25, opened 2020-12-15
Checksums


nevrome:
If the changes suggested in https://github.com/poseidon-framework/poseidon2-schema/issues/28 are introduced, we would have to update `trident`.

- `validate` should ensure that the checksums match
- `init` should create the checksums for new packages and add them to the POSEIDON.yml
- either `init` or a new module (`update`?) should update checksums in the POSEIDON.yml when the files were consciously changed

nevrome:
Implemented with #29.

Issue #24, opened 2020-12-12
Wrong YAML key order


nevrome:
The order of keys in the encoded POSEIDON.yml file is arbitrary and therefore confusing. It should be strictly defined.

The solution suggested [here](https://stackoverflow.com/questions/37865388/haskell-generating-json-with-aeson-gives-incorrect-order-of-fields) for Aeson did not work for me. Other solutions might build upon [Data.Yaml.Pretty](https://hackage.haskell.org/package/yaml-0.11.5.0/docs/Data-Yaml-Pretty.html) or [Data.Yaml.Builder](https://hackage.haskell.org/package/yaml-0.11.5.0/docs/Data-Yaml-Builder.html).

stschiff:
I would vote to not define the order. We deliberately chose YAML as the format for this file, and it feels to me like tacking on an order restriction will be something that we enforce outside of the format specification. It feels weird to me. Open to discussing, obviously, and don't feel strong about it, but I would say that I could live with the fact that some YAML files will have different order.

Nothing stops us from _recommending_ implicitly a specific order, simply by using a consistent order in all of our GitHub-backed files, as well as in our help messages. But I wouldn't enforce it

nevrome:
Right - ok - the wording above is misleading. The input order does not matter, but trident should always return files in a certain order. Just as it is now implemented in #29. Please check if you agree if this implementation.

stschiff:
I agree.

PR #29, opened 2021-01-09
Checksums

nevrome:
#25


nevrome:
Ah quick test with `data_ancient` showed 1. that the md5 sum calculation is fast enough to run like this (great!), but 2. `trident update` mixes file paths from different projects under certain (unknown) conditions. This is a crucial bug that has to be fixed before merging.

nevrome:
@stschiff 

So this is a large PR that addresses two main issues: #24 and #25. It also contains the changes in #28 (which fixes #23). I recommend to review #28 first. If there are any problems in #28 then it should be rejected and the issues should be fixed here in #29, if not, then #28 can be simply merged.

#23, #24 and #25 are IMHO the most important unsolved concerns before we can establish any stable and reliable poseidon repository. We need clear and detailed version management for each package (#23), writing POSEIDON.yml files should produce well ordered human-readable output (#24) and each data file should be clearly tracked with an md5 checksum (#25). 

This PR contains the necessary changes to do so::

- refactoring in `Package.hs` to handle the necessary backend operations as well as changes to the `PoseidonPackage` data type
- added `Checksum.hs` with functions for the checksum calculation and a dedicated data type `ChecksumListSpec`
- added `CLI/Update.hs` with the CLI module function `runUpdate`

A test run (within `data_ancient`) was successful: `trident update` adds correct checksums to each `POSEIDON.yml` file which are then correctly recognized (or rejected) by `trident validate`. `trident init` creates the new `POSEIDON.yml` file only with checksums for the genotype data (and not for `.janno` and `.bib` file) to reduce the mental burden on the end user.

Please test, review and (if you have the time) refactor carefully. I did not write any tests for these new features yet, but promise to do so eventually.

stschiff:
OK, we need to talk about this in person... I'm not convinced by the YAML order thing.

src-executables/Main.hs
```diff
@@ -103,6 +104,7 @@ fstatsOptParser = FstatsOptions <$> parseBasePaths
                                 <*> OP.many parseStatSpecsDirect
                                 <*> parseStatSpecsFromFile
                                 <*> parseRawOutput
+                                <*> parseIgnoreChecksums
```
stschiff, line 107: Hmm. Ist dieses Opt-out des Checksum-checks die beste Lösung? Sollten wir nicht einfach Warnings ausgeben wenn die Checksums nicht passen? Das sollte doch in der Regel schnell zu beheben sein, oder?

src/Poseidon/Package.hs
```diff
@@ -86,10 +99,37 @@ instance ToJSON PoseidonPackage where
...
+        "packageVersion" .= posPacPackageVersion x,
         "lastModified" .= posPacLastModified x,
         "bibFile" .= posPacBibFile x,
         "genotypeData" .= posPacGenotypeData x,
-        "jannoFile" .= posPacJannoFile x
+        "jannoFile" .= posPacJannoFile x,
+        "checksums" .= posPacChecksumList x
+        ]
+
+instance ToPrettyYaml PoseidonPackage where
```
stschiff, line 110: As far as I understand, we _can_ and should use the `ToPrettyYaml` type class to define field orders in all nested types within this Yaml structure... so also within `ContributorSpec` and `GenotypeSpec`

src/Poseidon/Package.hs
```diff
@@ -149,50 +189,118 @@ readPoseidonPackage yamlPath = do
...
         _ -> error "this should never happen")
-    subDirs     <- filterM doesDirectoryExist . map (baseDir </>) $ entries
-    morePosPacs <- fmap concat . mapM findPoseidonPackages $ subDirs
-    return $ (rights posPac) ++ morePosPacs
+    return $ uncurry zip $ second rights $ unzip $ filter (\(_, a) -> isRight a) pacsPacWithPaths
   where
     tryReadPoseidonPackage :: FilePath -> IO (Either PoseidonException PoseidonPackage)
     tryReadPoseidonPackage = try . readPoseidonPackage
 
+findAllPOSEIDONymlFiles :: FilePath -> IO [FilePath]
```
stschiff, line 210: I think we need to clean this up. There is a zoo of `loadXYZ`-related functions now...

src/Poseidon/CLI/FStats.hs
```diff
@@ -216,8 +218,10 @@ getPopIndices indEntries popSpec =
 
 -- | The main function running the FStats command.
 runFstats :: FstatsOptions -> IO ()
-runFstats (FstatsOptions baseDirs jackknifeMode exclusionList statSpecsDirect maybeStatSpecsFile rawOutput) = do
-    packages <- loadPoseidonPackages baseDirs
+runFstats (FstatsOptions baseDirs jackknifeMode exclusionList statSpecsDirect maybeStatSpecsFile rawOutput ignoreChecksums) = do
+    packages <- if ignoreChecksums 
+                then loadPoseidonPackagesIgnoreChecksums baseDirs
```
stschiff, line 223: I think it would be better if `ignoreChecksums` would be a simple argument to `loadPoseidonPackages`... that would already save one `load...` function

stschiff (commented):
OK, wow. Huge PR. I think it looks all great, I just have some issues with `Package.hs`, where I'd like to understand better the rationale behind some decisions. Let's aim for a meeting to discuss.

nevrome:
I now applied the necessary changes for my portion of #33. It was impossible for me to do this without some significant refactoring in the loading infrastructure and I'm curious what you think about the design choices I made, @stschiff.

I did not touch the whole context of relative and absolute paths, so there is still work left. For me it would be ok to merge before this is all sorted out, though. This PR is getting much too large.

stschiff (approved):
OK, looks all very good to me. OK to merge. I may still notice some things down the road, as always, but I don't want to stop this from being merged now. Thanks for the hard work!

Issue #38, opened 2021-02-04
trident fetch


nevrome:
We assume that trident will often be used in conjunction with remote data repositories. These should be servers that provide on request both janno packages with all data as well as a simple top-level information data structure in JSON format to document the names and version of the packages the respective repository holds. It should also store the endpoints where the data can be requested for download.

Trident should have a module called `fetch` to download information and data from these servers. The interface could look something like this:

```
trident fetch \
  -f "*pac1*,*pac2*" \ 
  --fetchFile path/to/entity/list/file \
  -d path/to/directory/where/pacs/should/be/stored \
  --preview \
  --remote https:://...
```

- `-f` defines which entities should be downloaded (see `trident forge`). At the beginning it will only support package names
- `-f` can also be given in a file `--fetchFile`
- For this module `-d` serves as the output directory. If multiple are given, then the first is used to store new data
- `--preview` only queries the info data a given server provides to show which packages are available there for download (similar to `trident list`)
- `--remote` allows to set a repo server URL. The default would be the DAG service

Fetch should automatically handle version update detection. That means if `-d` is not empty, the local and the remote package versions should be compared and only updated data should be downloaded. If `-f` and `--fetchFile` are empty, then this should be done for all packages in all `-d`s.

nevrome:
I just saw the nice download interface of singularity again. I think that's how I would like the fetch command line messages to look like:

```
Copying blob a4a2a29f9ba4 skipped: already exists  
Copying blob 127c9761dcba skipped: already exists  
Copying blob d13bf203e905 skipped: already exists  
Copying blob 4039240d2e0b skipped: already exists  
Copying blob 0288f53e6092 done  
Copying blob bfb2ff329e0c done  
Copying blob 0bca5c1eea79 [==================================>---] 229.9MiB / 248.2MiB
```

stschiff:
Yes, that looks very good!

Right now, the simplest would be to use the zip-files provided by the server, and as far as I can see, their size gets projected when - for example - downloaded via a browser. So a progress bar should in principle be possible.

We should also make `fetch` clever enough to check whether a download is necessary in the first place. 

And there should perhaps be an `--all` option, simply for users who know they want everything (like our department).

PR #47, opened 2021-02-11
Fetch

nevrome:
#38

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ A toolset to work with modular genotype databases formatted using Poseidon. The 
 * [Guide for the command line utility](#guide-for-the-command-line-utility)
   + [Poseidon package repositories](#poseidon-package-repositories)
   + [Analysing your own dataset outside of the main repository](#analysing-your-own-dataset-outside-of-the-main-repository)
-  + [Package creation and manipulation commands](#package-creation-and-manipulation-commands): [`init`](#init-command), [`forge`](#forge-command), [`update`](#update-command)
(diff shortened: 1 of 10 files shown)


stschiff:
Again impressive progress. Thanks. I'll checkout a local version and will have a look!

nevrome:
It's not ready yet, but feel free to investigate. Updating packages where older versions are present is not really handled so far.

nevrome:
So I think this generally works now. Needs testing and review :slightly_smiling_face:. Some open questions/ToDos:
- Right now only package names in the `forgeString` are considered, therefore the `forgeString` might not be the right choice here. Either this or we change the server code to also provide a list of individuals and groups.
- Related to that: The name `forgeString` is not really appropriate any more, now that we use it for both `forge` and `fetch`. If we keep the setup like this, then we should switch to a more generic name.
- I didn't write any user documentation for the README yet.

nevrome:
One more thing: Fetch does not include an option to see the remote server content yet (e.g. `--preview`). I'm also not sure if this should go into fetch. Maybe it's better to introduce yet another module like `preview` or integrate the remote listing with `list`.

stschiff:
OK, very impressive. Some comments/questions on the UI:
- I like the idea of putting the `--preview` functionality into `list`. It could be simply a `--remote` option in `list`, which would be a strict alternative to giving the base directories... you can either choose base directories or `--remote`.
- I think repurposing the forge-string is clever (apart from its name in the UI, indeed). How do you want the semantics to work here? When the user selects individuals or populations, I think ultimately we would still download the entire package, wouldn't we? We won't somehow try to download only those individuals, right?

I suggest you merge that PR in. I can't see any problems.

PR #48, opened 2021-02-14
Enable https

stschiff:
https support now implemented. When certification and key files (and optionally also chain files) are given, it serves via HTTPS.


stschiff:
@nevrome if you're OK, please pull this in. I would've pushed immediately into master, but was hesitating because of your large `fetch` PR. So as soon as you have merged fetch, please pull this one in. Thanks.

PR #55, opened 2021-02-20
More freedom for .janno files

nevrome:
As a first step I switched to `FromNamedRecord`. This is slower, but covers two out of the three requirements we had in mind for the new .janno file definition:

- [x] Ignoring additional columns beyond the ones in `PoseidonSample`
- [x] Arbitrary column order

Unsolved remains:

- [ ] Ignore missing columns (replace empty fields with `Nothing` in `PoseidonSample`)

Regarding this last point I posted [a question](https://stackoverflow.com/questions/66293020/haskell-cassava-data-csv-ignore-missing-columns-fields) on Stackoverflow, but I'm not positive that we will get an answer. So maybe it really has to be done with "manual" bytestring manipulation before reading them with cassava.


nevrome:
I did get an answer to my question and implemented the suggested solution. Indeed that works great: Missing columns are no issue any more. Unfortunately all other problems are silently ignored as well. Every table entry that can not be parsed is silently read as `Nothing`.

This could be considered a feature, because it makes reading .janno files ultra robust. Unfortunately it also breaks the file validation we had with the previous implementation. How could we make our .janno reading functions work with missing columns, but at the same time fail with wrong entries in individual cells? It must be possible, but I'm failing on the Applicative Functor syntax. Could you help me out, @stschiff?

stschiff (approved):
OK, fantastic. I love the Applicative approach using `FromNamedRecord`. I would approve this for now, I trust that you have tried it and it works. I would have to dive a lot deeper into Cassava to fully understand and evaluate everything here, but for now I'm content with seeing the general gist of it. Looks like very clean code to me!

stschiff:
I have no immediate answer to your question regarding missing vs. failing columns. I'll need to look deeper into this.

stschiff:
OK, I get the problem now... if you can't parse a field you just return Nothing... so parsing errors will be undetected... hmm

stschiff:
OK, this is definitely possible, but we can't work with `optional` here. I'll think of something.

stschiff:
OK, I've proposed the necessary change. It compiles, and tests pass, but I haven't yet added the necessary tests that check this new named-record parsing, so the specific thing I've just implemented doesn't really get tested.

nevrome:
Excellent! Thank you!

I will add some tests for this much more flexible janno file parsing.

Issue #53, opened 2021-02-17
Duplication check across packages


nevrome:
https://github.com/poseidon-framework/data_ancient/issues/5 reminded me that trident only checks for individual/sample duplicates on the  level of individual packages. But name duplicates across packages in a repo/baseDir may cause significant issues down the line - I just saw convertf fail because of that. 

On which level should we check for duplicates? And how should we report that? Warnings? Errors?

stschiff:
Hmm, at the very least, I think we should check for that in `forge`. Ideally even after loading packages from a base directory.

PR #56, opened 2021-02-20
Duplication check for individuals in collections

nevrome:
#53


stschiff (approved):
Great! Looks all good. If that increases latency we can still make it optional. But I anyway think the main bottleneck with running any trident command right now is the checksum-checking.

PR #57, opened 2021-02-25
implemented intersect option for forge.

stschiff:
New option `--intersect` for forge now allows to output only those SNPs that are common in all to-be-forged packages.

compiles and the output looks correct in terms of number of sites, but before merging we better make one actual test by a user.


stschiff:
OK, I'm gonna merge this in. If there is a bug, I'll fix it.

PR #58, opened 2021-02-26
Enable the list module to list from remote

### LINKED ISSUE #50: Make the list module capable of reporting not just from a local -d, but also from --remote
The interface and the output should be identical, except for `-d` <-> `--remote`.

nevrome:
If merged this closes #50. 

If you want to list local entities you run  

```
trident list -d . --individuals/--packages/--groups
```

So nothing changed here. For remote entities you replace `-d` with `--remote`:

```
trident list --remote --individuals/--packages/--groups
```

If both `-d` and `--remote` are provided, then -d is silently ignored. 

```
trident list -d . --remote --individuals/--packages/--groups
```
The remote's URL can be modified with `--remoteURL`.

### DIFF
--- a/src-executables/Main-trident.hs
+++ b/src-executables/Main-trident.hs
@@ -100,9 +100,9 @@ optParser = OP.subparser (
     initOptInfo = OP.info (OP.helper <*> (CmdInit <$> initOptParser))
         (OP.progDesc "Create a new Poseidon package from genotype data")
     listOptInfo = OP.info (OP.helper <*> (CmdList <$> listOptParser))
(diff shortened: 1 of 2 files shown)


stschiff:
OK, thanks for taking initiative. There are number of points I'd like to improve in the code. I'll make some commits.

stschiff:
OK, I've made some changes to simplify the code.

I think we can merge this. A question: Ist this a good time point to call this version 1.0.0? Or is there a major feature still missing?

nevrome:
Excellent - looks much cleaner now!

Issue #12, opened 2020-11-27
genoconvert module


nevrome:
Trident can deal equally well with plink binary and eigenstrat data. We could have a module that converts the genotype data in packages from one format to the other. That would also include replacing the format definition in the POSEIDON.yml file.

nevrome:
In this context `forge` should also get an option to define the desired output format

stschiff:
Indeed! However, my Plink module in `sequence-formats` doesn't feature Plink writing yet, just reading, so it'll have to wait until I get around implementing that.

stschiff:
OK, this is partly solved now with forge being able to output PLINK format.

nevrome:
Indeed `forge` covers that usecase then. But I think for ease of use we should add an explicit `convert` module. It would be a confusing user experience to go through `forge` if you just want to "quickly convert this package".

stschiff:
Hmm. What interface do you have in mind? Do you want to convert in place? That sounds dangerous... or do you want to just add the converted genotype files to the same package? That would be possible, but the new genotype datafiles would then not be covered by the `POSEIDON.yml` file (cause that can have only one genotype dataset)...

So I don't really see how this would be any different from a new forge command. With the interface I have in mind it'd be very similar:
```
trident convert -d basedir -n package_name --toEigenstrat -n new_pac_name -o /path/to/new/pac
```
versus
```
trident forge -d basedir -f "*package_name*" --eigenstrat -n new_pac_name -o /path/to/new/pac
```

or is there a cleverer, more intuitive interface?

nevrome:
Hm - I really imagined simple replacement. This conversion from EIGENSTRAT to PLINK and back is lossless, right? And you can only perform it, if you have writing permissions for the respective directories. So while I understand your concern, I also love how simple this would make things:

```
trident convert -d . --outFormat EIGENSTRAT
```

stschiff:
Hmm, OK, that's simple enough. I agree... and with `-d` in this case you would select a single directory? What happens if you run that on the entire repo? Does it then convert all packages?

nevrome:
It does. Probably we should ask the user for confirmation. 

And we should think carefully about the order of events in `convert` to prevent loss of data even in case of an interrupt: 
1. Conversion and therefore creation of new data files
2. Replacement of the paths in the POSEIDON.yml to point to the new data files
3. Finally: Delete the "old" data files

stschiff:
[...]

Issue #35, opened 2021-01-29
Update should also update the `lastModified` field


stschiff:
`update`, `forge` and `init` should have correct `lastModified` fields, as well as checksums

nevrome:
I implemented what I think is sensible in #29:

- `forge` and `init` do set the `lastModified` field to the current date, but do not add checksums as they are an advanced and optional feature mainly relevant for our central data archive.
- `update` does add/update checksums, but does not edit `lastModified`, because adding checksums is not per se a meaningful modification of the package

What do you think @stschiff?

stschiff:
I think forge should in fact add checksums. We need to make their check optional, but there is no good reason why a forged package shouldn't have them, in my view

nevrome:
You tend to be right about these things, but I'm not sure here. I fear the checksums will not be helpful, but annoying for users that are preparing their own datasets based on available packages with `forge`. 

I always envisioned the checksums to be only really relevant for our repository, where version-keeping is key. If we want to "force" them on the users for their own subsets, then we should have a good reason to do so. trident does not work with packages with broken checksums, so every tiny (!) change in the package files will require a subsequent call to `update` to bring the checksums back in sync! That will get frustrating soon if you're actively working with a dataset and not just preparing a static one for storage.

stschiff:
You make a good argument. OK, let's leave things as they are for now. We could mid-term then think about renaming `update`, which sounds awfully relevant, when it actually "only" computes the checksums, which you are right can be seen as a bit of an arcane job...

nevrome:
Renaming `update` is a good idea. Maybe `update-checksums` or simply `checksums`.

PR #62, opened 2021-03-07
genoconvert

nevrome:
As suggested and discussed already a loooooong time ago in #12.

This is a minimal implementation that just replaces the old genotype data with a converted version.

### DIFF
--- a/src-executables/Main-trident.hs
+++ b/src-executables/Main-trident.hs
@@ -166,7 +166,7 @@ genoconvertOptParser = GenoconvertOptions <$> parseBasePaths
                                           <*> parseRemoveOld
 
 parseRemoveOld :: OP.Parser Bool
-parseRemoveOld = OP.switch (OP.long "remove-old" <> OP.help "Remove the old genotype files when creating the new ones")
+parseRemoveOld = OP.switch (OP.long "removeOld" <> OP.help "Remove the old genotype files when creating the new ones")
 
 summariseOptParser :: OP.Parser SummariseOptions
 summariseOptParser = SummariseOptions <$> parseBasePaths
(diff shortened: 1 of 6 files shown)


stschiff:
OK great. I would only suggest to add one more option to control the removing-files behaviour. In fact, I would prefer to keep the old format files by default and only remove them if the user specifies that with `--remove-old` or so.
A second comment concerns the checksums. We currently don't create those, neither for `forge` nor for this new subcommand. Shouldn't we do that by default? I mean, when if not upon genotype-data creation should we compute checksums?

nevrome:
Thank you for checking!

Regarding keeping the old files: As I already stated [here](https://github.com/poseidon-framework/poseidon-hs/issues/12#issuecomment-786200604) I don't know then what should happen in case of a back-conversion. Should we overwrite then?

And regarding the checksums I made my point [here](https://github.com/poseidon-framework/poseidon-hs/issues/35#issuecomment-778679298): I think they are an advanced feature and only really relevant for our repository. We also loosened checksum tests to speed up trident, so many users might not get any value out of them. BUt maybe I'm missing an important usecase :man_shrugging: ?

stschiff:
Right, I think it's still the right decision not to check checksums by default, but they're definitely a plus. They give extra security that a package contains specific data. But I have to admit I also don't recognise an immediate benefit for the user... so I'm not strong about that.

With respect to switching to old data... I think we should simply print out an error if we want to write files that are already there. Frankly, if you already have genotype files in the other format, then you don't need genoconvert, you can just manually change the yaml file.

Let me propose that change as a commit and see how you like it.

stschiff:
OK, still need to put in a check for the output files.

src-executables/Main-trident.hs
```diff
@@ -163,6 +163,10 @@ forgeOptParser = ForgeOptions <$> parseBasePaths
 genoconvertOptParser :: OP.Parser GenoconvertOptions
 genoconvertOptParser = GenoconvertOptions <$> parseBasePaths
                                           <*> parseOutGenotypeFormat
+                                          <*> parseRemoveOld
+
+parseRemoveOld :: OP.Parser Bool
+parseRemoveOld = OP.switch (OP.long "remove-old" <> OP.help "Remove the old genotype files when creating the new ones")
```
nevrome, line 169: Afaik all our option names are in camelCase at the moment - so maybe `removeOld` would fit in better than `remove-old`

stschiff: Ah ok. Fine. I agree

stschiff (approved):

PR #66, opened 2021-03-12
Extend list command

stschiff:
the server now has a new API called /janno_all. This API is now used in the `list` command, which enables more flexible reporting of individual data. Specifically:

```
trident list --individuals -d /path/to/repo -j Country -j Data_Type
```

now lists - in addition to the default columns - also the country and the Data_Type from the janno files.

### DIFF
--- a/src-executables/Main-server.hs
+++ b/src-executables/Main-server.hs
@@ -198,14 +198,8 @@ makeZipArchive pac ignoreGenoFiles =
         let zipEntry = toEntry fn modTime raw
         return (addEntryToArchive zipEntry a)
 
-getAllIndividualInfo :: [PoseidonPackage] -> [IndividualInfo]
-getAllIndividualInfo packages = do
-    pac <- packages
-    jannoRow <- posPacJanno pac
-    let name = posSamIndividualID jannoRow
-        group = head . posSamGroupName $ jannoRow
-        pacName = posPacTitle pac
-    return $ IndividualInfo name group pacName
+getAllIndividualInfo :: [PoseidonPackage] -> [(String, Janno)]
+getAllIndividualInfo packages = [(posPacTitle pac, posPacJanno pac) |pac <- packages]
 
 optParserInfo :: OP.ParserInfo CommandLineOptions
 optParserInfo = OP.info (OP.helper <*> versionOption <*> optParser) (
--- a/src-executables/Main-server.hs
+++ b/src-executables/Main-server.hs
@@ -1,11 +1,10 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 import           Poseidon.GenotypeData       (GenotypeDataSpec (..))
-import           Poseidon.Janno              (PoseidonSample (..))
+import           Poseidon.Janno              (JannoRow (..))
 import           Poseidon.Package            (PackageInfo (..),
                                               PoseidonPackage (..),
                                               readPoseidonPackageCollection)
-import           Poseidon.Utils              (IndividualInfo (..))
 
 import           Codec.Archive.Zip           (Archive, addEntryToArchive,
                                               emptyArchive, fromArchive,
@@ -94,7 +93,7 @@ main = do
                 Just fn -> file fn
                 Nothing -> raise ("unknown package " <> p)
         get "/individuals_all" $
-            json (getAllIndividualInfo allPackages)
+            json (getAllPacJannoPairs allPackages)
         notFound $ raise "Unknown request"
   where
     p = OP.prefs OP.showHelpOnEmpty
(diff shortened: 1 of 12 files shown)


src/Poseidon/CLI/List.hs
```diff
@@ -42,53 +42,104 @@ runList (ListOptions repoLocation listEntity rawOutput ignoreGeno) = do
...
-            let tableH = ["Package", "Individual", "Group"]
-            let tableB = do
-                    oneSample <- allSampleInfo
-                    return [indInfoPacName oneSample, indInfoName oneSample, indInfoGroup oneSample]
+        ListIndividuals moreJannoColumns -> do
+            let tableH = ["Package", "Individual", "Group"] ++ moreJannoColumns
+            tableB <- fmap concat . forM allSampleInfo $ \(pacName, rows) ->
+                forM rows (\row -> do
+                    moreFields <- extractAdditionalFields row moreJannoColumns
+                    return ([pacName, jIndividualID row, head (jGroupName row)] ++ moreFields))
```
nevrome, line 78: Displaying only the first GroupName makes sense, but it may hide the other names in this column. Of course one could get the full column with `-j Group_Name`, but I wonder if this should not just be the default with `--individuals`. 

For `--groups` we have to limit this to the first group name, I guess. But maybe the help text should then mention this.

src-executables/Main-server.hs
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 import           Poseidon.GenotypeData       (GenotypeDataSpec (..))
-import           Poseidon.Janno              (PoseidonSample (..))
+import           Poseidon.Janno              (JannoRow (..))
```
nevrome, line 4: I love this more explicit name. I should have implemented it like this in the first place!

src/Poseidon/Utils.hs
```diff
@@ -64,7 +65,7 @@ renderPoseidonException PoseidonEmptyForgeException =
     "Nothing to be forged"
 renderPoseidonException (PoseidonNewPackageConstructionException s) =
     show s
-
+renderPoseidonException (PoseidonGenericException s) = show s
```
nevrome, line 68: The show here is superfluous, I believe. `s` is already a String. Wrapping it in `show` adds unnecessary quotes:
```
[clemens@cranberry data_ancient]$ trident list --individuals -d . -j Country -j GroupName --ignoreGeno
Searching POSEIDON.yml files... 134 found
Initializing packages... 134 initialized
Packages loaded: 134
Preparing output table
"GroupName is not a valid Janno column name"
```

src/Poseidon/Janno.hs
```diff
@@ -20,36 +19,39 @@ module Poseidon.Janno (
...
+                                             defaultOptions, genericToEncoding, toJSON, parseJSON,
+                                             toEncoding)
+import           Data.Bifunctor             (second)
 import qualified Data.ByteString.Char8      as Bchs
 import qualified Data.ByteString.Lazy.Char8 as Bch
 import           Data.Char                  (ord)
 import qualified Data.Csv                   as Csv
-import qualified Data.HashMap.Lazy          as HM
-import           Data.Either                (isRight, isLeft, rights, lefts)
+import           Data.Either                (isLeft, isRight, lefts, rights)
```
nevrome, line 37: Really? :laughing: 
Ah wait - your styler did this automatically?

nevrome (approved):
I tested the functionality and it worked exactly as described for me. All in all brilliant.

Also renaming the janno column variables must have been a lot of tedious work - thanks for that!

src/Poseidon/CLI/List.hs
stschiff: Yeah, I had thought about that too. Ideally we'd drop this behaviour of singling out the first group. In `list --packages`, we can just list all groups, in `list --groups` we would just list all the groups, even if some individuals appear in multiple groups`, and in `--individuals` there is no problem. For `list --groups` this would involve a slightly more involved fold. I'll submit a separate issue for that, as I think it's independent from this PR.

src-executables/Main-server.hs
stschiff: Yeah... I regretted it shortly after I did it :-) But I'm happy now.

src/Poseidon/Utils.hs
stschiff: Completely correct. I'll commit a patch

src/Poseidon/Janno.hs
stschiff: Yeah, that's just "stylish-haskell" in vscode.

nevrome: I just can't get this to work on my systems. Maybe I should investigate once more

Issue #64, opened 2021-03-11
do not check for duplicate individuals in inspection commands


stschiff:
We should not check for duplicate individuals - or at least not refuse to continue if it found some. It's easy to see that users have multiple package merges lying around in a directory, and in that case it's impossible to run `list` on that directory just to see what's there.

nevrome:
Right - I see that there might be some value in that. But I'm not convinced that this is a simple change:

- What should `list` display for the number of individuals in a group, if some individuals are there twice? 
- How should `summarise` deal with duplicates for counts and means?
- Which individuals should `forge` choose to merge?
- ...

Imho we could come up with a logic for list to handle that and only print a warning, but there are too many open questions to simply throw out the check at this point.

stschiff:
Sorry, I was too quick in my message. I definitely think for `forge` duplicate individual IDs should be forbidden. I'm specifically talking only about the "Inspection commands", so `list`, `summarise`, `survey` and `validate`. And I think ideally we'd print a quick warning of the sort "You have duplicate individuals in your packages: X, Y, Z..." (and printing only the first three that are duplicate". 

As to the behaviours, I think in `list` we should just print out the number of individuals including duplicates. In `summarise` also, just act as if all individuals are different.

nevrome:
Ah - ok. I misunderstood that. Makes sense like this. Let's implement it :+1:

PR #71, opened 2021-03-16
ignore duplicate individuals

nevrome:
This adds an option to `readPoseidonPackageCollection` to not halt on duplicate individuals. 

I activated it for `fetch`, `list`, `summarise` and `survey` as proposed in #64.

### DIFF
--- a/src/Poseidon/CLI/Checksumupdate.hs
+++ b/src/Poseidon/CLI/Checksumupdate.hs
@@ -1,5 +1,5 @@
-module Poseidon.CLI.Update (
-    runUpdate, UpdateOptions (..),
+module Poseidon.CLI.Checksumupdate (
+    runChecksumupdate, ChecksumupdateOptions (..),
     ) where
 
 import           Poseidon.Package           (PoseidonPackage (..),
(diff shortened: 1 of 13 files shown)


stschiff:
Damn it, I wanted to implement that. You were faster... low hanging fruit :-) OK, thanks, I'll go through this.

nevrome:
Scooping up the easy issues :laughing:

src/Poseidon/CLI/Checksumupdate.hs
```diff
@@ -0,0 +1,26 @@
+module Poseidon.CLI.Checksumupdate (
```
stschiff, line 1: What's the difference between this new CLI module and the former `update`?

stschiff (approved):
OK, haven't tried it, but looks all good.

src/Poseidon/CLI/Checksumupdate.hs
nevrome: Oh snap - that was accidentally copied from #70 :facepalm:. Maybe you could also give this other PR your blessing. Then I can merge them both at once. `checksumupdate` is just a clearer name for the former `update`.

Issue #77, opened 2021-03-25
looking for duplicate individuals in forge command


AyGhal:
suggestion: Maybe a better way to look for duplicates is only in package_names/inds/groups that are given to forge command not the entire base directories.

nevrome:
Implemented now.

Issue #76, opened 2021-03-25
Allow for multiple --forgeString and/or --forgeFile when using forge command


AyGhal:
as an enhancement to forge command when merging data from multiple directories.

nevrome:
Is implemented.

Issue #73, opened 2021-03-18
The POSEIDON.yml file fields readmeFile and changelogFile are dropped by the checksum update


nevrome:
They should be kept around if available.

PR #78, opened 2021-03-26
Multiple minor fixes

nevrome:
This is just a collection of small changes suggested in #73, #76 and #77. Everything backwards compatible.

@AyGhal had some good ideas to make working with `forge` a little bit more convenient :+1:


stschiff (approved):
OK, I've not tested stuff out in detail, but I trust that it works and I think then it's fine to merge.

Issue #72, opened 2021-03-18
Janno->JSON translation needs to respect text representation in Janno file


stschiff:
We currently have a few quirks in the JSON translation, such as "," instead of ";" as list separator, and "Modern" instead of "modern". We need to go through all these datatype ToJSON/FromJSON instances and make sure they're consistent with the text representation of Janno data as defined in the original janno files.

stschiff:
@nevrome do you think this is an actual bug? I don't think it is right? The way we convert from JSON and to JSON just needs to be unique and correct, but there is - other than consistency - no particular reason why we can't use different list separator symbols in the JANNO format and in JSON.

nevrome:
This issue is not a bug - it's an enhancement suggestion :stuck_out_tongue: 

But I guess your question is, if we really have to change it. I would suggest to do so to maintain consistency. Inconsistencies like this are often a long-term mental burden.

stschiff:
OK, of course, makes sense!

nevrome:
Turns out that there is not much mental burden after all. I close this now.

PR #81, opened 2021-04-15
added show instances for multiple janno column types

nevrome:
This is related to #72: To display the janno columns "correctly" (i.e. as in the janno files) in `list --individuals`, we need the right Show instance definitions (or something comparable). The JSON encoding is not relevant for that, although still a valid issue by itself.

This PR has some Show instances for our janno column types to make sure that the output of `list` and the representation in the janno files are identical. An issue arises for `Sex`, as it and its Show instance are defined in SequenceFormats.Eigenstrat and can not be overwritten here.

I assume a "clean" solution would be to switch to an own typeclass for janno-like printing (e.g. `JannoShow`), but that would mean going through the user facing code and replacing `show` with `jannoShow` wherever necessary.


stschiff (approved):
Looks great, thank you! I'll merge this because I'm preparing another PR.

stschiff:
Oops. Merge conflict. Can you take care of this, @nevrome?

nevrome:
Hm - the `sex` issue is not solved with this PR, unfortunately. But I will just open an issue for that.

nevrome:
Can be merged now

stschiff:
Great, waiting for the CI to finish, then I'll merge.

nevrome:
That will take more than an hour :shrug:

stschiff:
Oof. OK, will merge now.

PR #85, opened 2021-04-26
added simple 100-line running of genotype data.

stschiff:
Compiles, but not tested yet

### DIFF
--- a/src-executables/Main-server.hs
+++ b/src-executables/Main-server.hs
@@ -65,7 +65,7 @@ main = do
     updateGlobalLogger logger (setHandlers [fh] . setLevel INFO)
     infoM logger "Server starting up. Loading packages..."
     opts@(CommandLineOptions baseDirs port ignoreGenoFiles certFiles) <- OP.customExecParser p optParserInfo
-    allPackages <- readPoseidonPackageCollection True False ignoreGenoFiles baseDirs
+    allPackages <- readPoseidonPackageCollection True False ignoreGenoFiles True baseDirs
     infoM logger "Checking whether zip files are missing or outdated"
     zipDict <- if ignoreGenoFiles then return [] else forM allPackages (\pac -> do
         let fn = posPacBaseDir pac <.> "zip"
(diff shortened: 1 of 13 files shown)


stschiff:
OK, this seems to do what we want. Have a look please. This is what I get when I run it on the faulty Saag et al dataset:
```
cloud@poseidonserver2:~/dev/poseidon-hs$ trident validate -d ~/poseidon-repos/published_data/2021_Saag_EastEuropean
Searching POSEIDON.yml files... 1 found
Initializing packages... 1 initialized
Packages loaded: 1
Checking first 100 SNPs of genotype data
Error in the genotype data: "EigenstratSnpEntry {snpChrom = 1, snpPos = 776546, snpGeneticPos = 2.0242e-2, snpId = \"rs12124819\", snpRef = 'N', snpAlt = 'A'}: encountered illegal genotype Hom-Ref with Ref-Allele missing"
```

nevrome:
Alright - so I see two issues with this implementation:
- `validate` now fails without genotype data even with `--ignoreGeno`. We need it to run even without genotype data
- When validating a collection of packages the error message does not inform you in which one the error occurred

I think both of these issues could be solved by moving this check into `readPoseidonPackage` where all the other checks already live. There we already have the infrastructure to consider `--ignoreGeno` and are already working on the level of individual packages.

If you want to limit this check to `validate` then `readPoseidonPackage` and `readPoseidonPackageCollection` could get yet another flag for that. But it seems to be fast enough to run it also in `forge`, `genoconvert` and `fstats`.

stschiff:
OK, agreed. But I don't think we need to run anything special in `fstats` or `genoconvert`, where this is automatically happening through the standard forge mechanism.

nevrome:
Ah - ja - I get it. The question is though, if the pre-check would maybe yield a better error message even for `fstats` or `genoconvert`.

stschiff:
OK done. Have a look.

stschiff:
Ah sorry, haven't checked the tests. Just a sec.

stschiff:
OK, great. From my end ready to merge.

Issue #89, opened 2021-04-30
Changes for new Poseidon schema version


nevrome:
Consequence of: https://github.com/poseidon-framework/poseidon2-schema/pull/35

- [x] Add `snpSet` to the package and yml data types. (#90)
- [x] Define logic to create a correct `snpSet` entry for the new packages generated by `merge` and `init`. (#90)
- [x] Add `Genetic_Source_Accession_IDs` to the janno data type. String list column without special validation.
- [x] Add `Data_Preparation_Pipeline_URL` to the janno data type. Requires an URL data type with respective validation.
- [x] Make janno column `Publication_Status` a list column to allow for multiple values. This will have consequences for the bib file generation in `forge`.

nevrome:
All these ToDos are now covered in #90. So we can move the discussion there.

PR #90, opened 2021-04-30
snpSet

nevrome:
This PR should introduce the new POSEIDON.yml field `snpSet` (#89)

### DIFF
--- a/src/Poseidon/CLI/Init.hs
+++ b/src/Poseidon/CLI/Init.hs
@@ -2,7 +2,10 @@ module Poseidon.CLI.Init where
 
 import           Poseidon.BibFile           (writeBibTeXFile)
 import           Poseidon.GenotypeData      (GenotypeDataSpec (..),
-                                             GenotypeFormatSpec (..), loadIndividuals)
+                                             GenotypeFormatSpec (..), 
+                                             loadIndividuals,
+                                             SNPSetSpec (..)
+                                             )
 import           Poseidon.Janno             (createMinimalJanno, 
                                              writeJannoFile)
 import           Poseidon.Package           (PoseidonPackage (..),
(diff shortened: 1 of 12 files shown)


src/Poseidon/CLI/Init.hs
```diff
@@ -32,7 +35,7 @@ runInit (InitOptions format genoFile snpFile indFile outPath outName) = do
     let outInd = takeFileName indFile
         outSnp = takeFileName snpFile
         outGeno = takeFileName genoFile
-        genotypeData = GenotypeDataSpec format outGeno Nothing outSnp Nothing outInd Nothing
+        genotypeData = GenotypeDataSpec format outGeno Nothing outSnp Nothing outInd Nothing SNPSetOther
```
nevrome, line 38: Two options for `init`:

- We add an input field to let the user set the SNP set
- We find a clever way to determine the SNP set of the input genotype data automatically

src/Poseidon/GenotypeData.hs
```diff
@@ -92,6 +96,46 @@ instance ToJSON GenotypeFormatSpec where
...
+
+snpSetMerge :: SNPSetSpec -> SNPSetSpec -> Bool -> SNPSetSpec
+snpSetMerge SNPSet1240K         SNPSet1240K         _       = SNPSet1240K
+snpSetMerge SNPSetHumanOrigins  SNPSetHumanOrigins  _       = SNPSetHumanOrigins
+snpSetMerge SNPSetOther         _                   _       = SNPSetOther
+snpSetMerge _                   SNPSetOther         _       = SNPSetOther
+snpSetMerge SNPSet1240K         SNPSetHumanOrigins  True    = SNPSetHumanOrigins
+snpSetMerge SNPSetHumanOrigins  SNPSet1240K         True    = SNPSetHumanOrigins
+snpSetMerge SNPSet1240K         SNPSetHumanOrigins  False   = SNPSet1240K
+snpSetMerge SNPSetHumanOrigins  SNPSet1240K         False   = SNPSet1240K
```
nevrome, line 137: I did not test these two functions yet. But I'm a bit proud of them :smile:

nevrome:
I hope I considered everything now. There really is a lot to keep in mind even with such a simple change like adding a new janno column.

stschiff:
Thank you, awesome! I'll have a proper look. Might take until early next week.

src/Poseidon/CLI/Init.hs
stschiff: Hmm. I'm leaning towards having the user to input that via the command line.

Issue #84, opened 2021-04-26
JannoShow typeclass


nevrome:
To display the janno columns "correctly" (i.e. as in the janno files) in `list --individuals`, we need the right Show instance definitions (or something comparable).

#81 added some of the necessary code, but an issue arises for `Sex`, as it and its Show instance are defined in SequenceFormats.Eigenstrat and can not be overwritten here.

I assume a "clean" solution would be to switch to an own typeclass for janno-like printing (e.g. `JannoShow`), but that would mean going through the user facing code and replacing `show` with `jannoShow` wherever necessary.

I would like to do it to get a bit more familiar with custom typeclasses, but it probably does not have a high priority.

stschiff:
Wouldn't it be much simpler to just use a `newtype` around `Sex` and define a custom `Show` instance, and then use that newtype in the `JannoRow` struct?

stschiff:
So that would only mean you'd have to change the code that takes the sex data from the `fam` files and puts them into the Janno.

nevrome:
Hm - ok. Also true. But what about the points in the code where we compare different `Sex` instances from genotype data and janno files? Wouldn't it be more convenient to avoid such a wrapper newtype? Also I read that the output of `show` should theoretically be readable again by `read`. That would mean you should never even use `Show` for anything in production.

stschiff:
I'm confused. Where do we compare sex instances from genotype data and janno files? Also I don't get what you said about `read`...

nevrome:
Here for example: https://github.com/poseidon-framework/poseidon-hs/blob/c221f5e4c52dfc26a64020549a92182fdb19420d/src/Poseidon/Package.hs#L323

But ok - maybe that's not too frequent. And for the `Group_Name` we already have a type mismatch between geno and janno. So maybe it's indeed the way to go.

Here's what I mean regarding Show: https://stackoverflow.com/questions/32529524/what-is-the-correct-way-to-write-a-rich-haskell-show-function

> First of all, show is expected to produce a string following the Haskell syntax -- ideally, one should be able to cut & paste that into a Haskell file and have that work.

stschiff:
OK, yes, this comparison is easy to replace with some custom code comparing the Janno- with the Eigenstrat Sex version.

[...]

PR #91, opened 2021-05-08
Show JannoSex

nevrome:
This completes #81. I went for the simple solution and wrapped Sex in JannoSex to define a custom Show instance.

This renders #84 a nice-to-have for the far future.


stschiff (approved):
Well done! Haven't tested, but certainly looks good!

Issue #79, opened 2021-03-29
List non-standard .janno columns to identify non-breaking mistakes in the .janno file


nevrome:
https://github.com/poseidon-framework/published_data/pull/4 made me aware that it can be tricky to identify spelling mistakes in .janno file column names. 

Ideas to make this more easy:
- Print the non-standard column names while loading. Problem: This could become cumbersome and awfully verbose.
- Add the non-standard column names to the output of `survey`. Problem: This option might get overlooked easily.

The functionality could also go into `validate` :thinking:

stschiff:
How about we implement a simple `--strict` option for `validate` which fails if any of the standard janno columns are missing. We can debate whether it should also fail a janno-file has additional columns, which I think is fine, as long as they're truly extra columns and not wrongly typed standard columns. So I think simply checking and failing if any of the columns we recognise is missing would already help, wouldn't it?

nevrome:
I fear that would not be too useful. I assume usually you would have .janno files with missing columns, given the growing number of variables we're defining for that.

I think a `--verbose` switch for validate might be an option. Something that is forwarded to `readPoseidonPackageCollection` and triggers a more informative reading process.

PR #92, opened 2021-05-08
Verbose janno reading

nevrome:
This is my suggestion for a solution to #79. Just a `--verbose` switch in `validate`, that propagates down to `readJannoFile` and triggers some more console output about additional and missing columns there.

I'm not a big fan of a `--strict` mode, as laid out in #79. I think `verbose` is sufficient to spot these kind of mistakes quickly when preparing the package.


stschiff (approved):
Uff, I'm starting to question this concept of handing so much power down to `readPoseidonPackageCollection`. It's not the number of arguments I'm dreading, but the fact that every time we add a feature we have to go through all callers and add an argument... 

I wonder whether we should somehow allow for customisation of the package loading with callbacks or so... anyway. Certainly all great and looks good. Haven't tested it though

Issue #69, opened 2021-03-14
List-command shouldn't single out the first group name


stschiff:
We should make the listings fully symmetric with respect to multiple group names. This shouldn't be hard.

In `list --packages` and `list --individuals` there is no problem at all. In `list --groups` we need to implement a little fold that builds up a dictionary from group name to individuals, which would then mean that individuals with multiple group names would simply appear in multiple groups - just as one would expect it.

stschiff:
I have now marked this as bug, because it may result in unexpected behaviour if people rely on multiple group names

nevrome:
I feel like we already talked about this once, so sorry for asking again: The genotype data only has one group name. This group name has a special role and deserves special treatment. If somebody wants to access more group names with `trident list`, they have to add `-j Group_Name` to their individual-wise query. 

```
trident list --remote --individuals -j Group_Name
```

So how should the behavior sketched out above work in relation to this?

PR #93, opened 2021-05-08
An attempt to change the list logic to consider more group names

nevrome:
This could be a starting point to solve #69


stschiff (approved):
Oh wow, this is much simpler and pragmatic than what I had in mind, which would have involved a more complex fold over the package janno rows. This is great. I assume it works (haven't checked), but definitely looks good to me.

PR #98, opened 2021-05-14
Making snp set non mandatory

stschiff:


src/Poseidon/CLI/Forge.hs
```diff
@@ -90,7 +91,17 @@ runForge (ForgeOptions baseDirs entitiesDirect entitiesFile intersect outPath ou
...
+    let snpSetList = map (snpSet . posPacGenotypeData) relevantPackages
+    let packageNamesWithSnpSetOther =
+            map snd . filter ((==SNPSetOther) . fst) . zip snpSetList . map posPacTitle $ relevantPackages
+    when (not . null $ packageNamesWithSnpSetOther) $ do
+        hPutStrLn stderr ("A friendly warning: One or more of your source packages (" ++ show packageNamesWithSnpSetOther ++
+            ") has snpSet: Other, which will result in your forged package to also have \
+            \snpSet: Other. If this is intended, please ignore this message. If not, \
+            \please enter the correct snpSet in those packages to make sure our forging can deduce \
+            \the right target snpSet")
+    let newSNPSet = snpSetMergeList snpSetList intersect
```
nevrome, line 104: I guess you could just use the output of `snpSetMergeList` for this check, right? It already has all you need to decide if the warning should be shown. `packageNamesWithSnpSetOther` is not necessary.

src/Poseidon/GenotypeData.hs
```diff
@@ -60,7 +60,11 @@ instance FromJSON GenotypeDataSpec where
         <*> v .:? "snpFileChkSum"
         <*> v .: "indFile"
         <*> v .:? "indFileChkSum"
-        <*> v .: "snpSet"
+        <*> (convertMaybeSnpSet <$> (v .:? "snpSet"))
+      where
+        convertMaybeSnpSet :: Maybe SNPSetSpec -> SNPSetSpec
+        convertMaybeSnpSet Nothing = SNPSetOther
+        convertMaybeSnpSet (Just s) = s
```
nevrome, line 67: It certainly feels a bit hacky, but it's pragmatic :shrug:

nevrome (approved):

src/Poseidon/CLI/Forge.hs
stschiff: Well, I would definitely want to keep `snpSetMergeList` pure, so outputting error is not possible. Do you mean to use the result of that function to see whether it's `SNPSetOther`? Or do you mean I should add to that function to return also those input packages with `Other`? I would like to avoid making `snpSetMergeList` more complex, I like the simplicity.

src/Poseidon/GenotypeData.hs
stschiff: Oh no it's not hacky at all! On the JSON/YAML file, this is now formally a non-mandatory field, as described in the Schema, but on the internal data structure side, this will be a non-Maybe field. I think this is a nice and clean handler that mediates between these two worlds :-)

src/Poseidon/CLI/Forge.hs
nevrome: Yes - indeed. I meant instead of checking if there is any `SNPSetOther` in the list of packages, you can as well just check if `snpSetMergeList` returns `SNPSetOther`. It's simpler, faster and should yield an identical result.

src/Poseidon/GenotypeData.hs
nevrome: Alright, alright :smile: - just generally the two worlds should be as close to each other as possible.

src/Poseidon/CLI/Forge.hs
stschiff: yes, true. But for the error message to be useful, we do need the names of the culprits, so I'm afraid we still need this zipped `packageNamesWithSnpSetOther`, or do you have a better idea?

src/Poseidon/GenotypeData.hs
stschiff: Yes, that's true, and it's certainly a design choice that one can debate. An alternative would be to fully reflect to the non-mandatory nature also in our data structure and just _interpret_ it like `SnpSetOther` whenever it needs to, but that I also find a bit cumbersome.

stschiff: Hmm... actually now that I said it, I think that's better... perhaps I should make that change and see how it looks. Once cool advantage is that it would allow us to see, for example in `forge` whether there are package with `Nothing` in snpSet and make the warning even more explicit. I'll give it a go.

src/Poseidon/CLI/Forge.hs
nevrome: Ok - sorry. I missed the `show packageNamesWithSnpSetOther`

stschiff:
OK, further changes. I think it's now cleaner.

nevrome:
Excellent - good to go now! Don't forget to tag a new release :+1:

PR #102, opened 2021-05-27
Fix pedantic errors

stschiff:


stschiff:
Uff, this was quite a bit of stuff.

@nevrome please pay particular attention to some of the shadowing that I've resolved (I've added trailing underscores where a variable shadows one with the same name). There was one place with `loadedKB` or `loadedB` or so where I think there was a bug.

src/Poseidon/CLI/Fetch.hs
```diff
@@ -187,19 +188,19 @@ printDownloadProgress fileSizeMB = loop 0 0
             x <- await
             maybe (return ()) (showDownloaded fileSizeMB loadedB) x
             where
-                showDownloaded fileSizeMB loadedKB x = do
-                    let newLoadedB = loadedB + B.length x
+                showDownloaded fileSizeMB_ loadedB_ x = do
```
stschiff, line 191: Here and in the following: Please check @nevrome whether this is all correct.

nevrome: Good spotting! Not sure why that worked, but your solution works as well, so I will just accept that :shrug:.

nevrome:
Looks all fine to me - I also ran some tests and all still worked as expected

stschiff:
OK, I'll merge.

PR #105, opened 2021-06-03
Simple bibtex parser

stschiff:
This PR replaces the current heavy dependency on pandoc-citeproc and replaces it with some simple Parsec-based functions. I copied much of the code from the existing Haskell package "BibTeX" by Henning Thielemann, which is however dead. I put a comment into the code to make this clear.

The tests pass and also the validation of the existing published_data, so I think we're covered for now.

I don't think there is much to review in terms of code, but I nevertheless wanted to give you a chance to see it first, @nevrome.


nevrome:
I followed your advice and did not look at the code, but I did a forge test with 70 packages. Went through fine and produced a valid bibtex file. Excellent job!

It still removes duplicates when forging, right?

stschiff:
Yes, I have not changed the behaviour.

stschiff:
OK, I'll merge this and make a version bump in the last versioning number.

Issue #110, opened 2021-06-21
trident checksumupdate -> trident update


nevrome:
I'm thinking a lot about trident when showering. So this is my latest shower thought:
The ongoing discussion regarding the [poseidon format](https://github.com/poseidon-framework/poseidon2-schema) will most likely cause more changes to available packages. For our own convenience I would like to make `trident checksumupdate` more powerful and rename it to `trident update`.

Here's the interface I have in mind:

```bash
trident update -d . \
  --poseidonVersion "1.2.3" \
  --version "major" (or "minor" or "patch") \
  --checksumUpdate \
  --date "2023-05-17"
```

This should make it more easy to update a lot of packages a once. Especially the version incrementing would be extremely useful. I believe this would be fairly easy to implement, but really useful in the future.

Maybe we should also automatically create a CHANGELOG file - not so sure about that. Could be really useful in the long run, but is a bit more work.

nevrome:
I thought again today how extremely useful this would be to update packages. And we should ABSOLUTELY create and/or update the CHANGELOG file.

```
trident update -d . \
  --poseidonVersion "1.2.3" \ (default if nothing is selected: do nothing)
  --version "major" (or "minor" or "patch") \ (default: check if any checksum changed and if so update minor)
  --checksumUpdate \ (needs to be pared with a --ignoreGeno option)
  --date "2023-05-17" (default: today)
  --newContributor "Name,Email;Name,Email,..."
  --updateLog "I changed this and that: ..."
```

stschiff:
Hmm, OK. Late to the game. Definitely looks like a slick and useful interface! I guess I would have thought that just manually tweaking the YML file would be around the same amount of keyboard strokes, but of course it's nice to have this process of updating packages available in a somewhat more controlled way.

nevrome:
Hm - whenever we do somthing as simple as incrementing the `poseidonVersion`, we also have to increment the `packageVersion` and the `lastModified` date (just to make sure that fetch triggers a redownload for all of them). As every package has a different `packageVersion` and `lastModified` entry, that means individually changing 150+ files. So we're talking about a ton (!!) of keystrokes :stuck_out_tongue:

I would argue `update` is a very important key feature to keep our repository up-to-date.

stschiff:
[...]

PR #107, opened 2021-06-03
Golden tests

nevrome:
So this is an implementation of #101. Looks like a lot of code with many changes, but is actually a pretty simple setup. 

Look [here](https://github.com/poseidon-framework/poseidon-hs/tree/golden_tests#golden-tests) for a general introduction.

Some of the tests may be a bit pointless and some functionality is not tested at all. It's only a starting point to get some first coverage for our complex CLI functions.

PR #111, opened 2021-07-02
More general update module

nevrome:
First, fully untested draft to cover #110. I'll finish it asap


nevrome:
Here's the interface I came up with:

```
Usage: trident update (-d|--baseDir DIR) [--poseidonVersion ARG] 
                      [--versionComponent ARG] [--noChecksumUpdate] 
                      [--newContributors ARG] [--logText ARG] [--force]
  Update POSEIDON.yml files automatically

Available options:
  -h,--help                Show this help text
  -d,--baseDir DIR         a base directory to search for Poseidon Packages
                           (could be a Poseidon repository)
  --poseidonVersion ARG    Poseidon version the packages should be updated to:
                           e.g. "2.5.3" (default: Nothing)
  --versionComponent ARG   Part of the package version number in the
                           POSEIDON.yml file that should be updated: Major,
                           Minor or Patch (see
                           https://semver.org) (default: Patch)
  --noChecksumUpdate       Should update of checksums in the POSEIDON.yml file
                           be skipped
  --ignoreGeno             ignore SNP and GenoFile
  --newContributors ARG    Contributors to add to the POSEIDON.yml file in the
                           form "[Firstname Lastname](Email address);..."
  --logText ARG            Log text for this version jump in the CHANGELOG
                           file (default: "not specified")
  --force                  Normally the POSEIDON.yml files are only changed if
                           the poseidonVersion is adjusted or any of the
                           checksums change. With --force a package version
                           update can be triggered even if this is not the case.
```

Note the `--force` option to press an update even if no checksums changed.

One example: If I would like to add myself as a contributor for all packages I could run the following command

```bash
trident update -d . --newContributors "[Clemens Schmid](schmid@shh.mpg.de)" --logText "Added Clemens as a contributor" --force
```

This will modify the POSEIDON.yml files to 

- add the new contributor
- increment the package version number by one in the patch field
- update the lastModified field
- add `changelogFile: CHANGELOG.md` (if there was no CHANGELOG file before)

And it will create/modify the CHANGELOG file with the following line:

```
V X.X.X: Added Clemens as a contributor
```

where `X.X.X` is the respective new version number of the package

I did not write unit and golden tests yet, but I think you could try it anyway, @stschiff

stschiff:
Very nice. I haven't reviewed the code in detail, but I think I'm OK with merging this PR if the command works. Definitely very nice interface!

nevrome:
I feel bad about saying this, because I know that you/we have a lot of other things to do as well, but maybe you could at least run some test cases to check if things behave as expected. 

I also didn't write any tests yet. `update` might be a solution to fix the issue in #107 :thinking:

stschiff:
OK, will take a look. 

> update might be a solution to fix the issue in #107 🤔

How so?

Issue #83, opened 2021-04-26
`trident forge` should possibly alter the `Nr_autosomal_SNPs` entry in Janno files


stschiff:
`trident forge` can result in fewer SNPs listed in the target of the forge compared to the sources. This may particularly happen if you use the `--intersect` option. I would consider it currently a bug that this isn't done, as users clearly expect the number in `Nr_autosomal_SNPs` to be related to the genotype data, and not to some abstract original calling.

nevrome:
Ok - in a perfect world we would probably recount the number of non-missing SNPs at the end of the forge process, right? Or even offer an extra subcommand `updateJanno` to update the number. Is this what you had in mind?

stschiff:
Uff, yes of course. Goodness. This needs to be fixed! I'll work on it.

nevrome:
Thanks Stephan - though maybe we can merge #107 first. Probably more easy this way round.

PR #112, opened 2021-07-26
added a counting mechanism for SNPs after forging

nevrome:
This is an implementation of a simple non-missing SNP count in `forge` to solve #83.

Might be useful as an independent module as well - or as part of one of the inspection modules we already have in trident. 

Playing around with some packages in the published_data repo yielded small, but still curious deviations between this count and the numbers reported in the respective .janno files.


stschiff:
OK, thanks for doing my homework :-) I'll go through this ASAP.

stschiff:
Hmm. So if I understand correctly, your strategy here is to make a second pass through the joint genotype data after forging just to determine the number of non-missing SNPs? Isn't that exacerbating the performance problem? I mean... in principle it should be possible do to the forging and the SNP counting in one pass right?

stschiff:
In principle your code is definitely right. A pipe-fold is exactly what's needed for this. But I think it can be done within the main forge stream. I'll take a closer look and make a suggestion

stschiff:
I think we need to use the `Pipes.Prelude.tee` function (https://hackage.haskell.org/package/pipes-4.3.16/docs/Pipes-Prelude.html) for this. It can convert the final consumer in the forge loop into a pipe by forwarding everything. Then we can wrap all this in a fold. Needs a bit of thinking, but I think it's possible...

nevrome:
Right - I considered this as well, but then decided against it. I thought the SNP counting might be interesting also outside of forge, so I didn't want to entangle it too much with it. But I have no strong opinions here and certainly prefer a faster solution.

stschiff:
OK, I will take a stab at that when I'm done with the forge optimisation. I think we can keep it modular as a pipe component, but still run it inside a single pass.

stschiff:
Alright. I've managed to put the SNP counting into the main forge pipe (or rather, the other way around as it looks in the code). It's beautifully elegant to work with `P.tee` and `P.fold`, but I haven't really tested it yet.

src/Poseidon/CLI/Forge.hs
```diff
@@ -127,8 +124,33 @@ runForge (ForgeOptions baseDirs entitiesDirect entitiesFile intersect_ outPath o
...
+    hPutStrLn stderr "Creating .janno file"
+    let jannoRowsWithNewSNPNumbers = zipWith (\x y -> x {jNrAutosomalSNPs = Just y})
+                                             relevantJannoRows
+                                             (V.toList newNrAutosomalSNPs)
+    writeJannoFile (outPath </> outName <.> "janno") jannoRowsWithNewSNPNumbers
+
+sumNonMissingSNPs :: V.Vector Int -> (EigenstratSnpEntry, GenoLine) -> V.Vector Int
+sumNonMissingSNPs accumulator (_, geno) =
+    let nonMissingInt = V.map nonMissingToInt geno
+    in V.zipWith (+) accumulator nonMissingInt
```
stschiff, line 148: I've also wondered here in this function whether this may not result in a performance or space issue, since we're literally creating a vector for the garbage collector every single SNP. We need to check whether this makes things much slower... It is easy to use a mutable vector here, simply updating each slot in-space, using `IOVector` and `O.foldM`.

stschiff:
Could you perhaps merge master into this branch once we have merged #118? We can then directly check whether the SNP counting results in a performance issue, by comparing with our tests on real data.

stschiff:
OK, it's actually what I thought: This new version has a space leak (visible through top). We need to work on the fold.

stschiff:
Here is a possibly related space leak that I found some time ago with pipes and resource acquisition
https://github.com/Gabriel439/Haskell-Pipes-Library/issues/159

It's a bit different, but `runManaged` and `runSafeT` are tangentially related, I think.

stschiff:
I managed to reduce the memory leak by orders of magnitudes by using unboxed vectors for the counting. As a general rule, Unboxed Vectors should always be used if we're dealing with Ints or Doubles and the likes.

But I think the memory footprint still continuously increases now in the forge loop, even though it's now on the order to a megabyte or so every few seconds, so likely tolerable in the end. But it's still bugging me, of course.

I'll make one more attempt with IOVector, which might fix it.

stschiff:
If we can construct a minimal example with `Pipes.Safe` and `readFile` from `Pipes.Safe.Prelude` and then a fold which reproduces this leak, it would be awesome to put it on stack overflow or GitHub issue on the Pipes library...

stschiff:
OK, so I've now implemented this with `IOVector`, which is mutable and in this case the more reasonable choice. Unfortunately, there still is a mild memory leak... I observe 1Mb every two seconds in a large forge, starting with around 25Mb. This is not terrible, and I think in practice it won't affect runs much.

I think I would propose to wait for #118 and then merge master back into here to have the more optimal forging algorithm and then check again.

I will try to construct a more minimal example with a fold inside `SafeT IO` and will ask the pipes author how to avoid leaking there.

stschiff:
OK. I've merged master, and low and behold... the memory seems to max out at 33Mb right now. I can't observe an actual leak anymore. Could you perhaps replicate this, @nevrome? It's possible that the former leak was related to the former forge algorithm.

stschiff:
Ah, and I should say that I have _not_ actually tested the SNP counting yet. That would also be awesome if you could somehow actually try it, @nevrome, perhaps you had a test case from your original implementation?

nevrome:
Alright - it seems to work well and a test with and without `--intersect` yields an expected result. But as already pointed out above, in **all** my tests the count after forge deviated from the numbers in the .janno file by a few thousand SNPs. 

Where is this coming from? I think our counts here are reliable, right? It's stupidly simple, so I don't think there is an error in our code. But the values in the janno files should already be correct?

nevrome:
With 

```
trident forge -f "*2012_PattersonGenetics*,*2021_Wang_EastAsia*,*2021_GnecchiRuscone_KazakhSteppe*,*2020_Bergstrom_HGDP*,*2019_Flegontov_PalaeoEskimo*,*2019_Olalde_Iberia*,*2019_Narasimhan_centralSouthAsia*" -d . -o huhu -n huhu
```

I don't observe meaningful memory leaking. How could I reproduce your observation?

nevrome:
OK - nevermind. I went to f0e17001f33c9fbf9e520686bccd13381bc9a67d and saw it. Amazing! Seems to be fixed in the current version.

stschiff:
Ugh, @nevrome, help... can you please resolve that conflict? I'm not comfortable enough (yet) to mess with the Golden test files. Thanks!

nevrome:
Done

stschiff:
Excellent. Thanks. I'll merge this then.

PR #119, opened 2021-08-16
Minimal package template

nevrome:
I wanted to have `newMinimalPackageTemplate` for paagen and other derived applications. This was a good opportunity to also refactor `newPackageTemplate`.


src/Poseidon/Package.hs
```diff
@@ -475,43 +476,62 @@ getJointGenotypeData :: (MonadSafe m) => Bool -- ^ whether to show all warnings
...
     ,   posPacBibFileChkSum = Nothing
     ,   posPacReadmeFile = Nothing
     ,   posPacChangelogFile = Nothing
     ,   posPacDuplicate = 1
     }
 
+-- | A function to create a more complete POSEIDON package
+-- This will take only the filenames of the provided files, so it assumes that the files will be copied into
+-- the directory into which the YAML file will be written
+newPackageTemplate :: FilePath -> String -> GenotypeDataSpec -> Maybe (Either [EigenstratIndEntry] [JannoRow]) -> Maybe BibTeX -> IO PoseidonPackage
```
stschiff, line 505: Note that it should never be necessary to use `Maybe` together with List types, since a list already is an extended Maybe, with `[]` corresponding to `Nothing`. There is no semantic difference between `Nothing` and `[]`, so in fact such a type choice creates multiple options to express the same thing.

src/Poseidon/Package.hs
```diff
@@ -475,43 +476,62 @@ getJointGenotypeData :: (MonadSafe m) => Bool -- ^ whether to show all warnings
...
+newPackageTemplate baseDir name genoData indsOrJanno bib = do
+    (UTCTime today _) <- getCurrentTime
+    let minimalTemplate = newMinimalPackageTemplate baseDir name genoData
+        fluffedUpTemplate = minimalTemplate {
+            posPacDescription = Just "Empty package template. Please add a description"
+        ,   posPacPackageVersion = Just $ makeVersion [0, 1, 0]
+        ,   posPacLastModified = Just today
+        }
+    return $ completeBibSpec name bib $ completeJannoSpec name indsOrJanno fluffedUpTemplate
```
stschiff, line 514: matter of style, but I've made it a habit to avoid multiple uses of `$` in a single row. I never know in what order they'll be executed 🤷‍♀️ but maybe that's just me.

stschiff (commented):
Nice, just made minor comments.

src/Poseidon/Package.hs
nevrome: Alright - I replaced `Maybe Bibtex` with only `Bibtex`. Or did you mean I should also refactor `Maybe (Either [EigenstratIndEntry] [JannoRow])`? I was actually a bit proud of the neat pattern matching it allows then

src/Poseidon/Package.hs
nevrome: I replaced that :+1:

src/Poseidon/Package.hs
stschiff: Hmm. Yeah, OK, I get now what you're doing with that `Maybe Either` type. Well, for me it's OK to keep it, but I think I would have solved that slightly differently. I would have just taken `Either [EigenstratIndEntry] [JannoRow]` and then have both `Left []` and `Right []` being interpreted as `Nothing`. I find that slightly cleaner... but it's partly up to taste.

By the way, what happens if the user provides the wrong number of entries here? Is that checked or can that not happen because of upstream or downstream checks?

nevrome: Hm - I think I like my interface more. With your implementation the NULL case becomes slightly awkward, because you have to call the function with `Left []` or `Right []` when you actually want to express `Nothing`.

There are no number checks here. The function is used within forge and init and imho not vulnerable there, because the input is well prepared. But these functions are pretty useful also outside of trident, so maybe we should indeed add some more input checks :thinking:. Whenever I come across a case like that I remember Futharks [Size Types](https://futhark.readthedocs.io/en/latest/language-reference.html#size-types)!

stschiff: Yes all good then!

stschiff:
good to merge whenever you're ready

nevrome:
Ok - I came across some inconsistent behavior regarding the bibtex file in forge and init. In forge we only want a bibtex file, when the input packages come with bibliographic information. In init we always want it, to encourage adding bibliographic information to new packages (do we?). Both is not properly implemented.

So this is only partially related to `newPackageTemplate` and the changes in this PR, but still has to be fixed and we could very well do it here. 

The forge case is straightforward, but I'm not sure about init. Should we maybe pre-fill the .bib file with a dummy citation? 

```
@article{Einstein1935,
  title = {Can Quantum-Mechanical Description of Physical Reality Be Considered Complete?},
  author = {Einstein, A. and Podolsky, B. and Rosen, N.},
  journal = {Phys. Rev.},
  volume = {47},
  issue = {10},
  pages = {777--780},
  year = {1935},
  publisher = {American Physical Society},
  doi = {10.1103/PhysRev.47.777},
}
```

Or does this go to far?

stschiff:
I like that! And I like your Einstein-Reference, but I think it might be better to literally put dummy values, perhaps, to make this clearer.

nevrome:
Alright - I think I'm done (for now - see #122). What do you think about `dummyBibEntry`?

stschiff (approved):
Looks good.

PR #120, opened 2021-08-19
Improving fstats error out

### LINKED ISSUE #117: Too much debug info in `trident fstats` stdout
When using trident fstats currently, it prints to terminal information like which packages are used, and what statistics are being computed. 
```
Searching POSEIDON.yml files... 1 found
Initializing packages... 
> 1 
Packages loaded: 1
1 relevant packages for chosen statistics identified:
merged_hgdp_common_vars
Computing stats [...]
```
As well as chunk-wise calculations
```
computing chunk range (1,10120) - (1,31389837), size 30000, values [2.9604829624446615e-2,2.8189925420158246e-2,1.2122674970605535e-2,1.2225821024397388e-2,8.328896699420482e-3,1.2087044364659414e-2]
```
Before printing out the final results. 

The chunk-wise computations are apparently printed out in `stdout`, instead of `stderr`, so when the output of `trident fstats` is redirected into a file for saving (with `>`), a lot of log information makes it to the desired output file. This makes the output harder to parse/read/use.

Suggestion:
Print chunk-wise calculation info in `stderr` so they do not interfere with the parsing of the output, or create a separate output file for them.

stschiff:
Closes #117

### DIFF
--- a/src/Poseidon/CLI/FStats.hs
+++ b/src/Poseidon/CLI/FStats.hs
@@ -251,7 +251,7 @@ runFstats (FstatsOptions baseDirs jackknifeMode exclusionList statSpecsDirect ma
                 Left e  ->  throwM e
                 Right f -> return f
             let summaryStatsProd = purely folds statsFold eigenstratProdInChunks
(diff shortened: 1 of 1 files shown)


nevrome:
Looks fine to me - a simple fix

PR #126, opened 2021-09-17
Add CentOS built gh-action for conda deployment

stschiff:
Added a CentOS docker build workflow.

### DIFF
--- a/.github/workflows/Dockerfile.centos
+++ b/.github/workflows/Dockerfile.centos
@@ -1,0 +1,19 @@
+FROM centos:7
+
+RUN yum -y update
+RUN yum -y install zlib-devel wget ncurses-devel ncurses-compat-libs make gcc
+
+# Install GHC since stack's local install has issues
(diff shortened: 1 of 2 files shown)


stschiff:
@nevrome I don't know how to test workflows in non-master branches. Any clue?

nevrome:
Feel free to test it on the master branch - it doesn't break anything and to be honest I do that all the time for changes to CI actions :slightly_smiling_face: You could theoretically also change when the action runs in the `on:` section, but I wouldn't go there.

Maybe we could call the produced executable trident-conda or something like this. Otherwise I fear CentOS users would be tempted to use it, and I think they should not, right? It's builds on an old version of CentOS only relevant for conda, correct?

stschiff:
Right OK. Will merge. Hmm, it's CentOS7. And I agree it's misleading, because it makes no sense to call the other one `linux` then. Perhaps I'll call it `trident-conda-linux`.

stschiff:
But how do I test it? Can I create a new tag which I'll remove after testing?

nevrome:
Sure - just don't publish the release draft, if the build failed. You can delete a broken release draft and tag on github with the GUI and locally with `git tag -d v0.3.1`.

PR #127, opened 2021-09-17
Mini-PR for cleaning up stack file

stschiff:
I just cleaned up the stack.yaml file and just wanted to make that transparent, hence the PR. If one other person has seen this and agrees I'll pull it in.


nevrome (approved):
Looks fine to me - and as long as poseidon-hs is only developed by us two, I think you can push something like that directly to the master :shrug:. I trust the tests to catch issues with dependencies. 

Or would you like to be really strict about that? Then we should probably lock the master branch and prevent pushing to it entirely.

stschiff:
Yes, was a bit unnecessary... I just was reluctant to commit to Master while having two open PRs. but you're right, doesn't matter in this case.

Issue #122, opened 2021-08-20
`--minimal` flag for forge and init to reduce the minimal footprint of poseidon


nevrome:
I would like to introduce a flag `--minimal`/`--minimalPackage`/`--outMinimal` flag to all package creating modules. If this flag is activated, the return package should be created with `newMinimalPackageTemplate` instead of `newPackageTemplate` and thus skip all the context fluff (janno/bib etc.). 

The output package with this flag should only contain the genotype data and a POSEIDON.yml file (which could actually also be stripped of the now mandatory contributor field).

nevrome:
Implemented in v.0.22.0

PR #137, opened 2021-10-16
Minimal package option

nevrome:
This is an implementation of the idea presented in #122


stschiff (approved):
Looks very good. Good to merge! Thanks.

Issue #131, opened 2021-10-08
define snp set with bed file during forge


bvernot:
It would be great to allow the user to define a set of sites to keep when forging, that is neither the intersection or the union of the sites in a set of packages.

stschiff:
Yes, that would be no problem at all to implement... one would have to decide what to do in case there are sites requested that simply aren't in the dataset... mark them missing or simply skip, I guess those are the two options.

bvernot:
Hm yes, I'm not sure what would be best. If you mark them missing in the
whole dataset (and I suppose to set N N as the two alleles), then the
output would always be consistent and predictable. But maybe completely
missing sites would cause issues with some software?

nevrome:
Implemented in v0.23.0

PR #140, opened 2021-11-09
Allow custom SNP file in forge

stschiff:
I've started implementing the feature requested in #131.


nevrome:
Excellent! I'll run a test in a minute, but I'm sure it does what it is supposed to. Maybe @bvernot is interested in giving it a try as well. He was asking for the feature in #131, after all.

***

- The `LambdaCase` extension is cool! How do you discover these? I'm always overwhelmed when I look at the list of extensions I *could* use.
- I don't understand why the change to `test/testDat/testModules/ancient/Schiffels_2016` was necessary. You just removed an arbitrary snp from this small test package, right? This has the consequence that a dozen of derived files in the Golden test setup change as well. That's fine in general, but why? What do we win by changing all these files?

nevrome:
My test was successful. Brilliant! I got an irritating error message first, because my snp file for `--selectSnps` had a trailing linebreak. I guess that is a general issue when reading .snp files. Could `SequenceFormats.Eigenstrat` be taught to ignore empty lines? Or would that be harmful in an unexpected way? Should I open an issue [here](https://github.com/stschiff/sequence-formats)?

stschiff:
I will add one more check that the input SNP file is sorted, then prepare to merge.

bvernot:
This is great, thank you! I will put it through the paces soon.

stschiff:
OK, I've added one more test and check that the custom SNP file is ordered. If the CI testing goes through I'll merge this.

Issue #139, opened 2021-11-08
Trident forge stops working due to additional space in forge_file


neija2611:
The Trident forge run stopped without giving an error message at an additional space in my forge_file. No data from that point was further processed. I ended up with a package containing all expected information (.bib, .geno, .ind, .janno, .snp, .yml -files) but just for the populations included in my forge_file listed before the additional space.

stschiff:
Thanks for reporting. Will look into it

nevrome:
This throws a proper error now in v.0.24.0.

PR #143, opened 2021-11-11
Better entities parsing

nevrome:
I tried to improve the entities list parsing, not least in reaction to #139. I hope it works and fails (!) now as expected.


test/Poseidon/EntitiesListSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+    let g1 = "test/testDat/testEntityFiles/goodEntities1.txt"
+        g2 = "test/testDat/testEntityFiles/goodEntities2.txt"
+        b1 = "test/testDat/testEntityFiles/badEntities1.txt"
+    it "should parse good, single-value-per-line files correctly" $ do
+        g1res <- readEntitiesFromFile g1
+        g1res `shouldBe` [Ind "a", Group "b", Pac "c"]
+    it "should parse good, multi-value-per-line files correctly" $ do
+        g2res <- readEntitiesFromFile g2
+        g2res `shouldBe` [Ind "a1", Ind "a2", Group "b1", Pac "c1", Pac "c2", Group "b2", Group "b3"]
+    it "should fail to parse bad files and throw an exception" $ do
```
stschiff, line 48: Do you also have a test for the issue raised in #139, so an empty line in the forge file? I couldn't see it in the new files added with this PR.

stschiff (commented):
Very good to have all these tests. I made one comment...

test/Poseidon/EntitiesListSpec.hs
nevrome: The issue in #139 are additional spaces, not newlines. Empty lines are not allowed right now. I added a test case to reflect that. We could be more flexible here, of course :man_shrugging:. But I guess then we could also consider to add proper comments :thinking:

nevrome:
Alright - I added support for empty lines and comments. Could you please take another quick look, @stschiff? It was a great parsec learning opportunity (!!), but maybe my solution is not to your standards :sweat_smile:

Issue #115, opened 2021-08-03
forgeString should have an negation option to exclude entities


nevrome:
We already discussed this many moons ago and Margherita had a usecase today where this concept would have been useful. I imagine an interface like this:

`-f "*my_package1*,*my_package2*,!<that_one_individual_I_don't_want>,!that_group_I_don't_want"`

stschiff:
Yes, I think that would be very useful indeed! I was thinking about a similar option for `fstats` (although that functionality is anyway currently under question whether we want to keep it in trident).

So yes, I'm supportive of that (ignoring my bizarre Todo-list right now).

stschiff:
So this would be a feature for both `fetch` and `forge`, right?

nevrome:
Right - we could use the same parser for both. Although with `fetch` operating on a package level I don't see how the feature would be used there :thinking:

nevrome:
Implemented in v.0.24.0

PR #144, opened 2021-11-14
Negative selection

nevrome:
Yesterday night I woke up with the sudden realization that #143 is already very close to #115. So here's a PR on top of #143 that introduces exclusion to forgeFiles.

This introduces some more significant changes to the way we treat entities for forging and I had to make some implementation decisions (e.g. `EntitySign` and `SignedEntitiesList`). I'm sure you have some input here.

I did not document the changes so far to wait for your feedback, but I think from a user perspective all of this should be perfectly backwards-compatible. The tests may give you an idea how this looks and feels.


stschiff (commented):
OK, wow, thanks for the initiative. So, two main comments:

1) You've made a decision now that the order of entered entities does not matter. For example, `*Package1*, -GroupA, *Package2*` is the same as `*Package1*, *Package2*, -GroupA`. This matters, because if `GroupA` appears in both packages, the two would lead to different outcome when executed in order or not. You have basically decided that all includes come first, and then the excludes happen, right? I think that's OK, but it needs to be communicated clearly.

2) I think I'd like to propose a change to your types. I don't like your polymorphic `Include a` and `Exclude a` types. That system is too general, it would even allow something like `Include 5` and type-check. I think it would make a lot more sense to build that into the constructor. For example we could have something like 

```Haskell
data SignedEntity = Include Entity | Exclude Entity
```

But perhaps we should have a brief in-person chat about that as well.

nevrome:
Ok - point 2 is well taken. A polymorphic solution is not always preferable. Thanks for that explanation! I should change that.

Point one is a bit more tricky. Overloading the entity order with meaning could theoretically allow for even more complex forge operations, e.g. "give me the `GroupA` individuals from `Package2`, but not the ones from `Package1`". But I intuitively prefer the current logic. It's a lot simpler. Do we lose too much or is it good enough :thinking:

stschiff:
No, I think you're spot on with solution 1). I think it's actually good because it's very easy to understand and to communicate, we should just clarify right in the command line help exactly what's being done: First all positive entities are combined, then all negative ones applied.

nevrome:
Alright - I merge this into #143, fix the types and add some documentation :+1:

PR #145, opened 2021-11-17
Better errors janno parsing

nevrome:
I briefly looked into better parsing for .janno files and saw a trivial but significant option to improve the error messages.

With this patch, the issue where @TCLamnidis got stuck recently gets reported as:

```
Can't read sample in ./2010_RasmussenNature.janno in line 2: 
parse error (Failed reading: conversion error: Data_Type "1240k" not in [Shotgun, 1240K, OtherCapture, ReferenceGenome])
```

Much better than

```
Can't read sample in /mnt/archgen/users/lamnidis/2020_BarqueraCurrentBiology/2020_BarqueraCurrentBiology.janno in line 4: 
parse error (Failed reading: conversion error: empty) at ""
```

I think there is even more room for improvement without rewriting the whole parser. Cassava uses Attoparsec and should actually report the unconsumed leftovers. See [here](https://hackage.haskell.org/package/cassava-0.5.2.0/docs/src/Data.Csv.Encoding.html#decodeWithP'):

```haskell
decodeWithP' p s =
    case AL.parse p s of
      AL.Done _ v     -> Right v
      AL.Fail left _ msg -> Left errMsg
        where
          errMsg = "parse error (" ++ msg ++ ") at " ++
                   (if BL8.length left > 100
                    then (take 100 $ BL8.unpack left) ++ " (truncated)"
                    else show (BL8.unpack left))
```

But somehow this does not work for us and I failed to figure out why. So for the time being I resorted to removing the useless trailing `at ""`. But maybe you immediately see the issue here @stschiff.

### DIFF
--- a/src/Poseidon/Janno.hs
+++ b/src/Poseidon/Janno.hs
@@ -66,8 +66,8 @@ instance Csv.FromField JannoSex where
         | x == "F" = pure (JannoSex Female)
         | x == "M" = pure (JannoSex Male)
         | x == "U" = pure (JannoSex Unknown)
-        | otherwise = empty
-
+        | otherwise = fail $ "Sex " ++ show x ++ " not in [F, M, U]"
+ 
 instance Csv.ToField JannoSex where
     toField (JannoSex Female)  = "F"
     toField (JannoSex Male)    = "M"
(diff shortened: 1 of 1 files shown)


stschiff:
OK, thanks. I'd like to go through this in a quiet minute, will do so tomorrow likely.

stschiff (approved):
Excellent. Love your solution using `fail`! Perfect! I haven't checked it, but I trust you on that, feel free to merge this.

PR #151, opened 2022-01-06
Removed fstats command 

stschiff:
I have removed the fstats command, which is now part of `xerxes` of the `poseidon-analysis-hs` package. I have added a brief warning message to the basic CLI, so that users who might be looking for `fstats` can find a pointer. We can eventually remove that, of course.


nevrome (approved):
Excellent! I don't see any issues. Can be merged and the poseidon-hs version can go to 0.25.0, I guess. Is xerxes already functional? We should probably provide it on the cluster asap, in case somebody is already using `fstats` from my up-to-date `trident` version on the cluster :thinking:

Let's make this clear when we post our usual version update message in the slack/mattermost channel

stschiff:
Yes, `xerxes fstats` is functional, but still contains a massive memory leak, which might be OK for 1240K data and few stats, but is something that needs fixing. I think we can indeed merge this. I'll bump the version nr also.

Issue #154, opened 2022-01-24
forge command runs without package naming


neija2611:
While using the "forge" command I didn't specify a name for the new created Poseidon package. No error occurred and the files (.bim, .fam, .bed, .janno, .bib) were created anyhow, but without being named. The package was called "" in the .yml file.

nevrome:
Thanks for reporting that - it should indeed be impossible to create unnamed packages :+1:

esalmela:
I stumbled upon the same issue, but in a slightly different fashion: in trident forge, I did specify a name for the new package, with full path and everything, but erroneously with a '/' in the end. This produced the new folder I asked for, and an unnamed package within it as above (iirc).

Fwiw, I wonder what would have happened, had I similarly typoed an earlier trident init command. A symmetric possibility of error may exist there.

stschiff:
Ugh... I see. So we should i) check that the name is not empty, and ii) check that it results in valid filenames... thanks for reporting.

PR #158, opened 2022-01-30
added a check to fail on an missing output package name

nevrome:
#154


stschiff:
Thanks. On it

stschiff (approved):
Thank you. Perfect.

At some point we need to discuss whether this proliferation of PoseidonException constructors is still best practice. I have no prior experience to larger software systems, and I wonder whether there are other ways people keep Exceptions meaningful and on-point. But definitely this PR can be merged! Thanks.

Issue #161, opened 2022-02-05
EntitiesList semantics


stschiff:
We currently have the option to specify entity lists by including and excluding entities. This is currently used in `forge` and `fetch`, and it is also used outside, in `xerxes`. 

The current semantics (not part of the data types, but of the `forge` and `fetch` implementations) are, that all includes are executed, before any excludes. An alternative would be to execute includes and excludes in order. This has two advantages in my view:

1) Less obfuscated:
Currently, if the user species `a,-<b>,c,-<d>`, she needs to reorder these in her head to predict the result.

2) More power:
There is actually a good reason for executing these operations in order. For example:
`*PackageA*,-GroupB,<IndidivualC>` could be useful if `<IndividualC>` would be part of `GroupB`. In that case, I would want to exclude all individuals _but_ `<IndividualC>`. I'd argue this to be a common use case, and this would not be possible in the current semantics. Currently, you would first include `<IndividualC>` and then exclude it again by excluding entire `GroupB`, to which `<IndividualC>` belongs.

I think this would be a relatively small change, and I'm happy to take a stab. It can easily be implemented using recursion.

nevrome:
Argument 2 is a pretty strong one! So I guess we should indeed change this. Happy to be involved.

stschiff:
Good. I've already switched to `EntitiesList` in `xerxes` with the ordered semantics and happy to make this change for `fetch` and `forge`. Will submit a PR

nevrome:
To my understanding this is fully solved by #163. I close the issue.

PR #163, opened 2022-03-08
EntitiesList update

stschiff:
Adresses #161 and implements some stuff needed for `poseidon-analysis-hs`.


stschiff:
Alright, this has been a more substantial PR than I anticipated. I significantly changed the EntitiesList API. There is now a signed and an unsigned version, and the implementation uses a class to use polymorphic functions that work with both signed and unsigned lists. Signed Entities are important for `trident forge` and `xerxes fstats` and `xerxes ras`, but unsigned entities are important for `trident fetch`.

There is now a new feature in `trident fetch`, that you can also ask for groups and individuals, and the correct packages are being loaded.

nevrome:
Ok! Cool! This is a frighteningly big PR affecting the heart of trident. Before I start the review I have some general shower thoughts:

1. We should use this opportunity of a major change to  `forge` to replace the `--eigenstrat` flag with the more proper, flexible and future-proof `--outFormat` option used in `genoConvert`
2. This PR should be accompanied with a PR to the documentation/website repo to properly explain the new EntitiesList semantics
3. As this is a severe change that will potentially break the workflows of people using the current selection syntax (e.g. @neija2611) we should add a very obvious warning in case negative selection is used. Something like this:
> The semantics of --forgeString and --forgeFile have been changed in trident v0.X.X. Removing samples, groups or packages now follows a different logic. Please see the documentation here (http://...) to verify that your selection still behaves as you expect.

src/Poseidon/EntitiesList.hs
```diff
@@ -39,63 +40,98 @@ instance Show SignedEntity where
...
-    eL <- P.sepBy parsePoseidonEntity (P.char ',' <* P.spaces)
-    P.optional (P.many (P.char ' ') >> comment)
-    return eL
-
-parseSignedEntitiesListFromFile :: P.Parser SignedEntitiesList
-parseSignedEntitiesListFromFile = do
-    concat <$> P.sepBy (P.try emptyLine <|> comment <|> parseSignedEntitiesList) P.newline
+-- A class to generalise signed and unsigned Entity Lists. Both have the feature that they can be used to filter individuals.
+class EntitySpec a where
+    indInfoConformsToEntitySpec :: [a] -> IndividualInfo -> Bool
```
nevrome, line 45: This is not specifically for this line, but more of a general observation: The definition of `IndividualInfo` seems to be less powerful than it could be. 
```haskell
data IndividualInfo = IndividualInfo { 
      indInfoName :: String
    , indInfoGroups :: [String]
    , indInfoPacName :: String
    }
```
I don't understand your logic deep down, but I feel this could be more precise. Maybe we should introduce dedicated `type`s for these categories which could then also be used in `PoseidonEntity` :thinking: 
```haskell
type PacName = String
type GroupName = String
type IndName = String

data IndividualInfo = IndividualInfo { 
      indInfoName :: PacName
    , indInfoGroups :: [GroupName]
    , indInfoPacName :: IndName
    }
```
Maybe it's too painful, but `indInfoGroups` could be a [`NonEmpty`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-List-NonEmpty.html)?

src/Poseidon/EntitiesList.hs
```diff
@@ -39,63 +40,98 @@ instance Show SignedEntity where
...
+instance EntitySpec SignedEntity where
+    indInfoConformsToEntitySpec signedEntities (IndividualInfo indName groupNames pacName) = go signedEntities False
+      where
+        go [] r = r
+        go (Include (Ind   n):rest) r = if n ==     indName    then go rest True  else go rest r
+        go (Include (Group n):rest) r = if n `elem` groupNames then go rest True  else go rest r
+        go (Include (Pac   n):rest) r = if n ==     pacName    then go rest True  else go rest r
+        go (Exclude (Ind   n):rest) r = if n ==     indName    then go rest False else go rest r
+        go (Exclude (Group n):rest) r = if n `elem` groupNames then go rest False else go rest r
+        go (Exclude (Pac   n):rest) r = if n ==     pacName    then go rest False else go rest r
```
nevrome, line 58: I'm not sure about this recursive solution. Wouldn't a fold with a helper function for the comparison be more readable? Or do I miss something here?

nevrome (commented):
I found it difficult to review that. Hard to grasp the ramification of every change. I like a lot of the changes and left some minor comments on the aspects I don't.

I did not test the new build yet.

src/Poseidon/Package.hs
```diff
@@ -530,6 +526,15 @@ getJointGenotypeData showAllWarnings intersect pacs maybeSnpFile = do
...
         _ ->  return ()
 
+getJointJanno :: [PoseidonPackage] -> [JannoRow]
+getJointJanno = concatMap posPacJanno
+
+getJointIndividualInfo :: [PoseidonPackage] -> [IndividualInfo]
+getJointIndividualInfo packages = do
+    pac <- packages
+    jannoRow <- posPacJanno pac
+    return $ IndividualInfo (jPoseidonID jannoRow) ((getJannoList . jGroupName) jannoRow) (posPacTitle pac)
```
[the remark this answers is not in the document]
stschiff: I think multi-line list-comprehension (aka do-notation for lists) are one of the killer features in Haskell. I have actually started a blog-post about it ("List comprehensions on steroids"), because the only way to come close to that in python or R would be to define a function and then map it over the list, which is arguably less readable. Furthermore, this is even a nested-for-loop, with an external and internal loop, and the do-notation automatically collapses them to a one-dimensional output. That's not even possible in python or R just like that... I don't think there is a better or more succinct way of writing that. I'm a huge fan of multi-line list comprehensions in Haskell!

BUT: As one of only two Haskell-developers in this project you have a big say here, and I'm very open to suggestions to make the code more readable :-)

src/Poseidon/CLI/Forge.hs
```diff
@@ -67,44 +72,49 @@ pacReadOpts = defaultPackageReadOptions {
...
 runForge :: ForgeOptions -> IO ()
 runForge (ForgeOptions baseDirs entitiesDirect entitiesFile intersect_ outPath maybeOutName outFormat minimal showWarnings noExtract maybeSnpFile) = do
+    
     -- compile entities
     entitiesFromFile <- mapM readEntitiesFromFile entitiesFile
-    let entities = nub $ entitiesDirect ++ concat entitiesFromFile
-        entitiesToIncludePreliminary = entityIncludes entities
-        entitiesToExclude = entityExcludes entities
+    let entitiesInput = nub $ entitiesDirect ++ concat entitiesFromFile
+        entitiesToInclude = [e | Include e <- entitiesInput]
```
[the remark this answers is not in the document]
stschiff: No! Here we want to include only the `Include`s. `removeEntitySign` returns all entities, even if it's an `Exclude`

src/Poseidon/EntitiesList.hs
```diff
@@ -39,63 +40,98 @@ instance Show SignedEntity where
...
+        parsePac   = Pac   <$> P.between (P.char '*') (P.char '*') parseName
+        parseGroup = Group <$> parseName
+        parseInd   = Ind   <$> P.between (P.char '<') (P.char '>') parseName
+        parseName  = P.many1 (P.satisfy (\c -> not (isSpace c || c `elem` ",<>*")))
+
+removeEntitySign :: SignedEntity -> PoseidonEntity
+removeEntitySign (Include e) = e
+removeEntitySign (Exclude e) = e
+
+entitiesListP :: (EntitySpec a) => P.Parser [a]
```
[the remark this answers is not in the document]
stschiff: Yeah, no... I don't think we're very consistent here... I started it in this module, because I got confused previously with the mixed naming. Yes, I'm also a fan of making these things more consistent.

src/Poseidon/EntitiesList.hs
stschiff: Hmm, but `type` is simply an Alias, not really a type. It doesn't provide any additional compiler-constraints. For example, even with these three type annotations, you can still do:
```
indName :: String
indName = "foo"
groupNames :: [String]
groupNames = ["group1", "group2"]
pacName :: String
pacName = "bar"
```

So, if you really want to make that more type-safe, you'd have to use a `newtype` or `data` declaration, which then would literally forbid supplying strings. I'm not sure whether that's the way to go. Semantically a group-name is literally a name, so a string, there is no additional semantic quality to it which would warrant a new-type, in my view.

But `type` only makes it look nicer, with no additional compiler guarantees whatsoever, I'm afraid, so I don't think that will help much.

stschiff: And regarding `NonEmpty`: Well, I'm curious how that would go. My understanding so far of `NonEmpty` is that while it's very nice in principle, it would have a pretty substantial cost downstream, because basically none of the typical list methods like concat and stuff would work. All these would have to replaced. Maybe that's not too hard to do, but remember that it would mean also changing this in the Janno definition, and therefore it would show up everywhere in the code-base. I'm not against it, just haven't done any research how this works.

src/Poseidon/EntitiesList.hs
stschiff: Yeah, I guess a `fold` could also work, but I find recursions to be very natural in Haskell now, and I learned thinking in terms of them more easily than in `fold`s. I basically wouldn't be able to cast this into a fold easily, I believe. Happy for suggestions, though.

stschiff:
Yes, I apologise for the substantial changes. I think it may be easiest to just go through `Forge` and `Fetch`, and then go top-down. 

I also think that in cases such as this, our Golden Tests are just awesome. They are the true guarantee that we haven't screwed something up.

src/Poseidon/EntitiesList.hs
stschiff: Hmm, very interesting, this `NonEmpty` suggestion. I just realised that that type is part of `base` and seems to support most important type classes (`Functor`, `Foldable`...), and all the necessary functions are there.

I think that's a very good suggestion. But I'm a bit anxious that it would require quite some code changes, since it would be already be part of Janno...

stschiff:
OK, there is another problem I've realised. Right now, entities are read both from the CLI (`entitiesDirect`) and given file-paths. Implicitly, they are read in order
```
    entitiesFromFile <- mapM readEntitiesFromFile entitiesFile
    let entitiesInput = nub $ entitiesDirect ++ concat entitiesFromFile
```

But that of course may have unintended consequences given the new order-dependency. 

My radical suggestion: Simplify the CLI-API to _either_ an ordered signed entity list given via the command line, _or_ via a single file. So do not allow multiple files, do not allow both CLI and file input. In my view, there is hardly a use-case for either multiple entity files or both CLI input and files, so I don't think this will harm anyone.

I'm happy to make that change, but wanted to quickly check with you.

stschiff:
(and I've also now changed the behaviour of `forge` as we've discussed, if the list starts with an exclude)

nevrome:
> My radical suggestion: Simplify the CLI-API to either an ordered signed entity list given via the command line, or via a single file. So do not allow multiple files, do not allow both CLI and file input. In my view, there is hardly a use-case for either multiple entity files or both CLI input and files, so I don't think this will harm anyone.

Ja - ok. I see how this can easily become confusing. Go for it. :+1:

src/Poseidon/Package.hs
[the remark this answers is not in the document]
nevrome: Alright - let's keep it then. List comprehensions exist in Python and R, btw.. No idea if they are this elegant, though.

src/Poseidon/CLI/Forge.hs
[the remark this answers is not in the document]
nevrome: I see. I still have difficulties with the syntax for generators in list comprehensions. It's like a language in a language. I didn't understand the code correctly.

src/Poseidon/EntitiesList.hs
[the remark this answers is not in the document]
nevrome: At some point we should do some major refactoring to clean up all this heterogeneity. poseidon-hs is probably pretty complete in terms of features. So we could think about cleaning and shrinking the code base.

src/Poseidon/EntitiesList.hs
nevrome: Thanks for the explanation about `type`. That was a misconception of mine. Only really discovered `type` recently.

I'll open an issue to consider `NonEmpty` in future refactoring. Doesn't have to be part of this PR.

src/Poseidon/EntitiesList.hs
nevrome: I implemented my solution in #165.

nevrome:
Alright - I'm waiting for your OK now to run some practical tests with the new interface.

nevrome:
Ok - I ran some tests and I think I found somethings that may be a bug - or at least unexpected behaviour. I tested this with a clone of the published_data repository.

This works fine:

```
$ trident forge -d . --forgeString "*2014_RaghavanNature*" -o huhu
...
Forging with the following entity-list: [*2014_RaghavanNature*]
Searching POSEIDON.yml files... 151 found
Checking Poseidon versions... 
Initializing packages... 
> 151 
Packages loaded: 151
1 packages contain data for this forging operation
...
```

But excluding the same package has an unforeseen effect:

```
$ trident forge -d . --forgeString "-*2014_RaghavanNature*" -o huhu
...
Forging with the following entity-list: [-*2014_RaghavanNature*]
Searching POSEIDON.yml files... 151 found
Checking Poseidon versions... 
Initializing packages... 
> 151 
Packages loaded: 151
forge entities begin with exclude or are empty, so implicitly adding all packages as includes before applying excludes.
The following entities do not exist in this dataset and will be ignored: *2014_RaghavanNature*
150 packages contain data for this forging operation
...
```

So we seem to get the right selection, but with a confusing error message:
> The following entities do not exist in this dataset and will be ignored: *2014_RaghavanNature*

When I first read this, I was convinced I must have made a typo in the package name.

This seems to happen only if the first entry in the list is an `Exclude Pac`.

nevrome:
When practically working with this version, I also find the `Forging with the following entity-list: ...` output pretty annoying. I have 3000 individuals in my forgeFile, so I'm presented with a wall of text. One way to solve this would be to truncate after 3-10 entities.

nevrome:
When practically working with this version, I also find the `Forging with the following entity-list: ...` output pretty annoying. I have 3000 individuals in my forgeFile, so I'm presented with a wall of text. One way to solve this would be to truncate after 3-10 entities.

stschiff:
All good points. I'm on it.

stschiff:
I think it should be fixed now.

nevrome:
Safe for a small adjustment of the documentation on the website this good to go, imho.

stschiff:
OK, I'll merge this then and adjust the documentation

PR #165, opened 2022-03-11
indInfoConformsToEntitySpec

nevrome:
An alternative implementation for `indInfoConformsToEntitySpec` for `SignedEntity`. I tested it with ghci and the (golden) tests run through as well, so I think it should be fine and behave as expected.

If you prefer your implementation, then we can decide with a fair coin toss, @stschiff :smile:


nevrome:
This fails now, but that's because of the changes in [pimp_entitieslist](https://github.com/poseidon-framework/poseidon-hs/tree/pimp_entitieslist). I'll merge this now.

Issue #150, opened 2021-12-18
Support more genotype data formats


nevrome:
The library could support some more genotype data formats (see e.g. [here](https://reich.hms.harvard.edu/software/InputFileFormats))
- [ ] ~ANCESTRYMAP~
- [x] EIGENSTRAT
- [ ] PED
- [x] PACKEDPED
- [ ] ~PACKEDANCESTRYMAP~
- [ ] VCF (see also https://github.com/poseidon-framework/poseidon2-schema/issues/45)
- [ ] ... ?

Would of course also require changes in the Poseidon schema.

stschiff:
Hmm... I'm not sure with respect to ANCESTRYMAP and PACKEDANCESTRYMAP. I have never seen them elsewhere at all. I think in Poseidon we should stick to relatively well understood and widely used formats... of course, we could think about making it easier for users to pull in datasets from different formats, but I think in the package we should save them in Eigenstrat or Plink, as we had it, and maybe in the future VCF (although this will be a bit of a rabbit hole, because VCF allows for _so much_ richer data than the other two formats...)

nevrome:
Adding new formats is a discussion that should start in the schema repository. And we already have an issue for that there: https://github.com/poseidon-framework/poseidon2-schema/issues/45

I'll close this one here now.

PR #168, opened 2022-03-30
Drill out forge/genocovert cli interface

nevrome:
Alright - so here's a draft for a change that has been on my mind for quite some time now: `trident forge` and `trident genoconvert` are powerful tools beyond Poseidon. And there is demand for a simple interface for operations on genotype data. So I hereby suggest to make `forge` and `genoconvert` capable of also handling "unpackaged" (in a Poseidon sense) data.

Under the hood this works by wrapping the input data in "pseudo" packages with the new `makePseudoPackageFromInGenotypeData :: InGenotypeData -> IO PoseidonPackage`, so that it can be treated with all the neat facilities we already established for proper Poseidon packages. Instead of asking the user to run `trident init`, we do it internally ourselves, when `forge` or `genoconvert` are called.

Now for the interesting part: The interface. I wanted this to be as flexible as possible, so I didn't go with a `Either [FilePath] InGenotypeData`, but decided to allow both kinds of input simultaneously. Here's a valid call to `genoconvert` making use of this flexibility:

```bash
trident genoconvert \
  -d 2015_LlorenteScience \
  -d 2015_FuNature \
  -r PLINK -g 2018_Mittnik_Baltic/Mittnik_Baltic.bed -s 2018_Mittnik_Baltic/Mittnik_Baltic.bim -i 2018_Mittnik_Baltic/Mittnik_Baltic.fam \
  -r PLINK -g 2010_RasmussenNature/2010_RasmussenNature.bed -i 2010_RasmussenNature/2010_RasmussenNature.fam -s 2010_RasmussenNature/2010_RasmussenNature.bim \
  --outFormat EIGENSTRAT
```

As before these changes, this will convert the genotype data in `2015_LlorenteScience` and `2015_FuNature`. But it will also convert `2018_Mittnik_Baltic/Mittnik_Baltic.bed/.bim/.fam` and `2010_RasmussenNature/2010_RasmussenNature.bed/.bim/.fam`, which I treat here as unpackaged genotype data.

Running `forge` with this system allows for even more fun applications:

```bash
trident forge \
  -d 2017_GonzalesFortesCurrentBiology \
  -r PLINK -g 2017_HaberAJHG/2017_HaberAJHG.bed -s 2017_HaberAJHG/2017_HaberAJHG.bim -i 2017_HaberAJHG/2017_HaberAJHG.fam \
  -r PLINK -g 2018_VeeramahPNAS/2018_VeeramahPNAS.bed -i 2018_VeeramahPNAS/2018_VeeramahPNAS.fam -s 2018_VeeramahPNAS/2018_VeeramahPNAS.bim \
  -f "<STR241.SG>,<ERS1790729.SG>,Iberia_HG.SG" 
  -o huhu
```

Here we forge across one Poseidon package, but also two unpackaged datasets, and take individuals/groups from all of them for our output package. Of course we can work across only real packages or only unpackaged datasets.

I also added an option `--onlyGeno` to both `genoconvert` and `forge`, to further cut down the output beyond `--minimal` and return nothing but the (unpackaged) genotype data files. That means a potential user can make use of these beautiful subcommands, without ever touching the Poseidon infrastructure. That is imho also very useful for automatic pipelines, where `--minimal` currently still forces you to drag along a POSEIDON.yml file. I was tempted to call `--onlyGeno` `--superMinimal`.

I'm aware that is a controversial change with many potential pitfalls. Let me know what you think, @stschiff. The interface should be fully backwards compatible, so this isn't a breaking change.


stschiff:
Beautiful interface and addition, need to review the code still!

stschiff:
> Of course we can work across only real packages or only unpackaged datasets.
What does this mean?

src/Poseidon/GenotypeData.hs
```diff
@@ -25,6 +25,16 @@ import           SequenceFormats.Plink      (readFamFile, readPlink)
 import           System.Console.ANSI        (hClearLine, hSetCursorColumn)
 import           System.FilePath            ((</>))
 import           System.IO                  (hFlush, hPutStr, hPutStrLn, stderr)
+
+-- | A simplified datatype to specify genotype files for ínput purposes
+data InGenotypeData = InGenotypeData
```
stschiff, line 30: Isn't this almost 100% code duplication from `GenotypeDataSpec`?

src/Poseidon/Package.hs
```diff
@@ -576,6 +578,23 @@ newMinimalPackageTemplate baseDir name (GenotypeDataSpec format_ geno _ snp _ in
     ,   posPacDuplicate = 1
     }
 
+makePseudoPackageFromInGenotypeData :: InGenotypeData -> IO PoseidonPackage
+makePseudoPackageFromInGenotypeData (InGenotypeData format_ genoFile_ snpFile_ indFile_ snpSet_) = do
+    let genotypeData = GenotypeDataSpec format_ genoFile_ Nothing snpFile_ Nothing indFile_ Nothing (Just snpSet_)
+        baseDir = getBaseDir genoFile_ snpFile_ indFile_
+        pacName = takeBaseName genoFile_
```
stschiff, line 585: OK, I like this idea to just take the base name of the geno-file for the package name

src/Poseidon/Package.hs
```diff
@@ -576,6 +578,23 @@ newMinimalPackageTemplate baseDir name (GenotypeDataSpec format_ geno _ snp _ in
...
+    newPackageTemplate baseDir pacName genotypeData (Just (Left inds)) []
+    where
+        getBaseDir :: FilePath -> FilePath -> FilePath -> FilePath
+        getBaseDir g s i =
+            let baseDirGeno = takeDirectory genoFile_
+                baseDirSnp = takeDirectory snpFile_
+                baseDirInd = takeDirectory indFile_
+            in if baseDirGeno == baseDirSnp && baseDirSnp == baseDirInd
+               then baseDirGeno
+               else throwM $ PoseidonUnequalBaseDirException g s i
```
stschiff, line 596: I think this is actually an unnecessary restriction. Of course it will rarely matter, since most users would indeed keep the three files in one directory, but we really don't need this requirement at all. The `baseDir` is just a root directory   off of which the individual file names are relative. So instead of searching for a common base-directory of these three files, we could simply take the current working directory as the root directory, right? The user specifies the three files in either absolute or relative paths, but if they are relative, they are meant to be relative from the current working directory, so we can just take that as the baseDir. We _should_ however check that the directories given by the user aren't absolute, in which case we'd probably have to take care that either we set `/` as the baseDir, or we check that our path logic still works with absolute paths if we stick with the CWD as the baseDir.

Am I making sense?

stschiff: (my point is: since we never will write out that package, the baseDir here is just a search path...)

src/Poseidon/CLI/Forge.hs
```diff
@@ -48,15 +50,17 @@ import           System.IO                   (hPutStrLn, stderr)
...
     { _forgeBaseDirs     :: [FilePath]
+    , _forgeInGenos      :: [InGenotypeData]
     , _forgeEntitySpec   :: Either SignedEntitiesList FilePath
+    , _forgeSnpFile      :: Maybe FilePath
     , _forgeIntersect    :: Bool
-    , _forgeOutPacPath   :: FilePath
-    , _forgeOutPacName   :: Maybe String
     , _forgeOutFormat    :: GenotypeFormatSpec
     , _forgeOutMinimal   :: Bool
+    , _forgeOutOnlyGeno  :: Bool
```
stschiff, line 59: As I know you like types reflecting program correctness: Here you could tie `onlyGeno` with `outName` in a clever way. Since we know that only package-output will require a name, you could define a datatype `data ForgeOutputSpec = ForgeOutputPackage String | ForgeOutputOnlyGeno`, which then forces the CLI to ensure that you can't do something like `--onlyGeno --pacName XXX`.

src/Poseidon/CLI/Genoconvert.hs
```diff
@@ -64,7 +72,7 @@ convertGenoTo outFormat removeOld pac = do
         let [outG, outS, outI] = map (posPacBaseDir pac </>) [outGeno, outSnp, outInd]
         anyExists <- or <$> mapM checkFile [outG, outS, outI]
         if anyExists
-        then hPutStrLn stderr ("skipping genotype convertion for package " ++ posPacTitle pac)
+        then hPutStrLn stderr ("skipping genotype convertion for " ++ posPacTitle pac)
```
stschiff, line 75: typo `conversion`

src/Poseidon/CLI/Init.hs
```diff
@@ -1,10 +1,9 @@
 module Poseidon.CLI.Init where
 
 import           Poseidon.BibFile           (dummyBibEntry, writeBibTeXFile)
-import           Poseidon.GenotypeData      (GenotypeDataSpec (..),
-                                             GenotypeFormatSpec (..), 
+import           Poseidon.GenotypeData      (InGenotypeData (..),
```
stschiff, line 4: See my other comment... I think these are redundant data types. They should be merged.

src/Poseidon/Utils.hs
```diff
@@ -31,6 +31,7 @@ data PoseidonException =
     | PoseidonRemoteJSONParsingException String -- ^ An exception to indicate failed remote info JSON parsing
     | PoseidonGenericException String -- ^ A catch-all for any other type of exception
     | PoseidonEmptyOutPacNameException -- ^ An exception to throw if the output package lacks a name
+    | PoseidonUnequalBaseDirException FilePath FilePath FilePath -- ^ An exception to throw if genotype data files don't share a common base directory
```
stschiff, line 34: as discussed, don't think we need this.

stschiff (commented):
Alright. Sorry I've opted for "Add single comment` when I should've put all these under the umbrella of my review. Looks very good conceptually... implementation-wise I left some comments.

src/Poseidon/GenotypeData.hs
nevrome: Yeah - I started with that, but then ran into issues when writing `parseInGenotypeDataset :: OP.Parser GenotypeDataSpec`, because the input does not have checksums. I then also though it might be conceptually useful to keep the input perspective separate. 

How would I write a parser for `GenotypeDataSpec` that sets the checksums to `Nothing`?

src/Poseidon/Package.hs
nevrome: Ok - that was maybe unnecessarily restrictive. c57e2b83df52c663c666fadebcd2a82ad11bcdd3 replaces it with a system that simply takes all relative paths

src/Poseidon/Utils.hs
nevrome: Let's discuss this above. I tried to replace it, but discovered a reason why to enforce this.

src/Poseidon/CLI/Init.hs
nevrome: Let's discuss this above.

src/Poseidon/CLI/Forge.hs
nevrome: I appreciate the idea, but as the connection between `onlyGeno` and `pacName` is a bit indirect, I think I prefer too keep things simple here. A `pacPath` is necessary anyway and the name is just a `Maybe`...

src/Poseidon/Package.hs
nevrome: So to explain this a bit more: For genoconvert we need a clear basedir to write to, because where else would you write? `.` doesn't make sense.

Maybe we should add an explicit output directory option for genoconvert, but I want to keep the default to write where the other data already is. That's just convenient. A user-specified output dir would have to be provided for every input package/pseudopackage. A lot of typing for little gain, so even if we add this, it should be optional.

Another option would be to write to the location of one or multiple of the input files. But that's also not very elegant and might become confusing easily.

I would argue, sometimes restrictions make a program more easy to use.

stschiff: Hmm, not following 100%. There is a clear expectation from the user that the new genotype files will be written exactly to where the old ones reside. So the idea here is to take whatever the input paths are and simply change the extension from `.bed` to `.geno` and from `.bim` to `.snp` and so on. That should then preserve the relative paths given to the programs. In 99,9% of cases, users will simply have the three files in one place, so this will all be the same to your restriction. In the rare case where some users might have the three files spread out over several directories, our program would simply follow that and place the new three files alongside the old ones. 

I see no problem.

src/Poseidon/GenotypeData.hs
stschiff: Very easy using applicative syntax and `pure Nothing`. I can write a more concrete code snippet later.

src/Poseidon/Package.hs
nevrome: I considered this, but not all genotype data formats consist of these three files. This will fall apart if/when we come back to #150

src/Poseidon/GenotypeData.hs
nevrome: I think I understand. I read a blog post about applicative yesterday. The `pure` was missing in my first attempt.

stschiff: Right. Like this:

```
parseInGenotypeDataset :: OP.Parser InGenotypeData
parseInGenotypeDataset = InGenotypeData <$> parseInGenotypeFormat
                                        <*> parseInGenoFile
                                        <*> pure Nothing
                                        <*> parseInSnpFile
                                        <*> pure Nothing
                                        <*> parseInIndFile
                                        <*> pure Nothing
                                        <*> parseGenotypeSNPSet
```

src/Poseidon/Package.hs
stschiff: OK fine then :-) We're discussing a marginal thing, I admit. I think most people won't even feel this "restriction". So fine to go ahead with the restriction.

src/Poseidon/GenotypeData.hs
nevrome: I removed `InGenotypeData` :+1:

PR #170, opened 2022-04-14
changed CLI interface for forgeString

stschiff:
very simple change that 
1) allows `forge` to work without either a forgeString or a forgeFile, interpreting as simply as `-f ""`, so an empty list of forge entities triggering a total merge.
2) removes the possibility to give multiple forge-strings, which I don't think makes much sense given that we already parse comma-separated values.


nevrome:
Would be OK for me :+1:

PR #172, opened 2022-05-09
Simpler geno interface

### LINKED ISSUE #171: update request for trident forge
Can we expand the new option for forge so that we can specify only the prefix for cases where all of our files (geno/snp/ind) are in the same directory and separately for cases where they are in different directories?

nevrome:
See `CHANGELOGRELEASE.md` for a summary of the changes in this PR. If/When this is merged it essentially closes #171

### DIFF
--- a/CHANGELOGRELEASE.md
+++ b/CHANGELOGRELEASE.md
@@ -1,5 +1,26 @@
 ### V ...
 
+### V 0.29.0
+
+This release brings two additions to the interface. They should make it more easy to work with unpackaged genotype files.
+
+`trident genoconvert` gets the option `-o/--outPackagePath`, which allows to redirect the conversion output to any directory. If this option is used and if the input data is not read from unpackaged genotype data, then the POSEIDON.yml file of the source package is not updated.
+
+The second, more significant change is an additional interface option to input unpackaged genotype data. This affects the trident subcommands `init`, `genoconvert` and `forge`. Besides the verbose interface with `-r + -g + -s + -i`, it is now also possible to only give `-p/--genoOne` to fully describe one unpackaged genotype data set. `-p` takes one of the genotype data files (so `.bed`, `.bim` or `.fam` for PLINK or `.geno`, `.snp` or `.ind` for EIGENSTRAT) and determines based on its extension the data format (PLINK/EIGENSTRAT) and the paths to the other files forming the dataset (assuming they have the same name and are in the same directory).
+
+Coming back to the `forge` example below for `V 0.28.0`, we can now for example write:
+
+```
+trident forge \
+  -d 2017_GonzalesFortesCurrentBiology \
+  -r PLINK -g 2017_HaberAJHG/2017_HaberAJHG.bed -s 2017_HaberAJHG/2017_HaberAJHG.bim -i 2017_HaberAJHG/2017_HaberAJHG.fam \
+  -p 2018_VeeramahPNAS/2018_VeeramahPNAS.fam \
+  -f "<STR241.SG>,<ERS1790729.SG>,Iberia_HG.SG" \
+  -o testpackage
+```
+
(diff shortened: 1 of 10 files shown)


nevrome:
@stschiff Because you interacted so intensely with #173: Did you see this more concrete PR #172? I think this one could actually be merged in the near future and I was hoping for your review.

CHANGELOGRELEASE.md
```diff
@@ -1,5 +1,26 @@
...
+
+The second, more significant change is an additional interface option to input unpackaged genotype data. This affects the trident subcommands `init`, `genoconvert` and `forge`. Besides the verbose interface with `-r + -g + -s + -i`, it is now also possible to only give `-p/--genoOne` to fully describe one unpackaged genotype data set. `-p` takes one of the genotype data files (so `.bed`, `.bim` or `.fam` for PLINK or `.geno`, `.snp` or `.ind` for EIGENSTRAT) and determines based on its extension the data format (PLINK/EIGENSTRAT) and the paths to the other files forming the dataset (assuming they have the same name and are in the same directory).
+
+Coming back to the `forge` example below for `V 0.28.0`, we can now for example write:
+
+```
+trident forge \
+  -d 2017_GonzalesFortesCurrentBiology \
+  -r PLINK -g 2017_HaberAJHG/2017_HaberAJHG.bed -s 2017_HaberAJHG/2017_HaberAJHG.bim -i 2017_HaberAJHG/2017_HaberAJHG.fam \
```
stschiff, line 16: As an example this still looks a bit scary for a Changelog. I would suggest you also use the new `-p` feature in the first package case, so `-p 2017_HaberAJHG/2017_HaberAJHG.bed`.

stschiff (commented):
OK, very nice. I've left one comment for the Changelog. On a more general note for the API. I personally think I'd prefer a system where you just give the basename of the genotype files without the extension, and then have the program look for extensions that match Eigenstrat or Plink and convert to the other format. There are two reasons for this:
1) Having the user choose one of the files is confusing, I'm sure we'll have users think that it HAS to be the fam file because that's in the help or so... it would just be cleaner to have them give just the basename.
2) There is no universally accepted file ending system for Plink or Eigenstrat. Some people like `geno.txt`, `.snp.txt` for Eigenstrat data, and for Plink it doesn't have to be `bed` and `bim` but sometimes other formats, too, look for example into the convertf program in Eigensoft. For now, of course, we have to assume these endings to be following the clear rules we currently use, but in the future we might do format detection using the content of the files rather than endings. Choosing now an API that is a bit more independent of the endings (for the input format) might ease the transition to this. 

What do you think?

CHANGELOGRELEASE.md
stschiff: and for the record, following my more general API comment, this would then be `-p 2017_HaberAJHG/2017_HaberAJHG`

nevrome:
As already discussed offline I'm not against a prefix based system, but I see technical and conceptual issues with that. The most obvious being, that the file type detection would obviously be an IO action, which can not run within the optparse-applicative parsers. We would have to build a way around that. Not impossible, but most likely not as elegant as the current interface:

```haskell
parseInGenotypeDataset :: OP.Parser GenotypeDataSpec
parseInGenotypeDataset = createGeno <$> (parseInGenoOne <|> parseInGenoSep) <*> parseGenotypeSNPSet
```

CHANGELOGRELEASE.md
nevrome: Yes - that's a good point :+1:

stschiff:
Hmm, OK... The only chance to save the prefix-idea would then be to relegate the detection logic to the main program and have the CLI simply record the user wish... so instead of parsing directly into `GenotypeDataSpec` from the CLI, parse first into some kind of `data GenotypeCLIspec = GenotypeCLIone String | GenotypeCLItriple String FilePath FilePath FilePath` (where the first parameter in the second case would be the type). We would then have a function `getGenotypeSpec :: GenotypeCLIspec -> IO GenotypeDataSpec` to be called in `main`. 

Anyway, I admit it's perhaps not worth it, and we can just go with what we have.

stschiff:
So I'm OK with merging that in!

Issue #51, opened 2021-02-16
use `hslogger` for all stderr output


stschiff:
As done now in `poseidon-http-server` I suggest to switch all stderr output in the core `poseidon-hs` library to logging functions à la `infoM` or `debugM` and the like (https://hackage.haskell.org/package/hslogger-1.3.1.0/docs/System-Log-Logger.html).

This has the advantage that we can control the output from outside the library, using functions like `updateGlobalLogger` or `trapLogging`.

As a first step, we could simply replace all `hPutStrLn stderr` with `infoM`, and then add `updateGlobalLogger rootLoggerName (setLevel INFO)` to the beginning of the `main` functions in any clients. This would then literally keep behaviour exactly the same for the user. As a next step, we could use different levels, for example `debugM` for debug-messages, or `warningM` for warning messages, which would then mean we control the output with verbosity levels even from the commandline. 

Using `updateGlobalLogger` we can even channel output into files or the syslog (could be interesting for the server). 

Using `traplogging` we can control how uncaught exceptions are output. For example using the ALERT level. 

Note that all this is a "nice-to-have", not a bug-fix.

nevrome:
To my minimal understanding `hslogger` operates fully in IO, right? Shouldn't we aspire to a more flexible and a bit more pure solution like the ReaderT pattern? I think it's sort of an industry standard. Maybe these could even be combined? But my googling revealed that for simple logging, hslogger is probably the most simple and clear solution out there.

About the ReaderT pattern: https://www.fpcomplete.com/blog/2017/06/readert-design-pattern
A simple example: https://stackoverflow.com/questions/61780295/readert-design-pattern-parametrize-the-environment

nevrome:
There are also other logging libraries. `co-log` looks promising: https://kowainik.github.io/posts/2018-09-25-co-log

nevrome:
Implemented now via co-log in #173

PR #173, opened 2022-05-09
Co-log experiment

nevrome:
Alright - so this is an experiment to tackle #51. I wanted to dive into the magic world of monads and monad transformers - and logging seems to be one common problem solved with these mighty tools.

I used [co-log](https://github.com/co-log/co-log) and an (I think) pretty simple setup. This is not a full implementation, but just a proof of concept. It compiles, though.

The core part is this, which now lives in `Poseidon.Utils`:

```haskell
type PoseidonLogIO = LoggerT Message IO

usePoseidonLogger :: PoseidonLogIO a -> IO a
usePoseidonLogger = usingLoggerT logAction
    where
        logAction :: LogAction IO Message
        logAction = cmap formatLogMessage logTextStdout
        formatLogMessage :: Message -> Text
        formatLogMessage me = showSeverity (msgSeverity me) <> msgText me
```

If a function yields any (log) messages then it should return not `IO ()`, but `PoseidonLogIO ()`. Then one can use all of these nice log message producing functions like `logWarning`, `logInfo` etc. in it. `LoggerT` is an instance of `MonadIO`, so one can still run IO actions within these functions - unfortunately always with this ugly `liftIO`, but I understood that many Haskellers are willing to accept that.

In this proof of concept `validate` yields the following output:
image
This looks neat and is a lot more easy to read. It would be easy to add more magic here: e.g. timestamps for each log message.

Let me know what you think! Is this too complicated? Or not complicated enough, so e.g. not general or future-proof enough?


stschiff:
Hmm, OK, I have to think about this. But basically this means that every IO action that needs logging needs to be augmented to live in the new Monad Stack. I'm not sure whether that is necessarily better than just logging directly from IO, which I think requires less of a change. But I need to read that blog post on co-log to understand its advantages.

stschiff:
OK... I went through the code an I'm actually very impressed. This looks pretty advanced and cool stuff! I tried to read through the blog post you've mentioned in #51 and found that my intellectual capabilities were inadequate... although the fact that I read it 11pm at night before going to bed might have contributed to my confusion. I will definitely now go through this again and try to get a hang of it. I like to understand better what the advantages are over the - arguably much simpler - hslogger within IO. But I'm definitely supportive of a more advanced technology here...

stschiff:
I think I'm sold, in principle, but perhaps still good to talk in person what exactly that change would entail for the rest of the code-base

nevrome:
I think this could (!) work like this. The colog logging is now implemented for all the subcommands of trident. I did not touch the web server, because I guess you would want to use the opportunity to write an own logging action that writes to a nice log file, @stschiff. `simpleLog` and `tridentDefaultLog` are probably not suitable for a server.

***

My biggest struggle is, that we "log" warnings extremely deep in `getConsensusSnpEntry`, whose logging behavior can not easily be controlled from the top layer. Right now it's hardcoded:

https://github.com/poseidon-framework/poseidon-hs/blob/27c21bcae8db60d1a3a7539a5a6dde73ce920e7b/src/Poseidon/GenotypeData.hs#L186-L187

I see three options to solve this:

1. Solve it properly with a monad stack compatible with `PoseidonLogIO`, that goes to the very depths of poseidon-hs. All my efforts to do that have been [in vain](https://github.com/poseidon-framework/poseidon-hs/tree/logInTheDeep). An extremely frustrating experience, although valuable for didactic reasons. I'm sure you could pull that off, @stschiff, but I'm not sure if we even want that.
2. Pass the `LogModus` down, just as we pass the Boolean for `Print Warning: yes/no`. That's a bit ugly, but clear and simple.
3. Pass the warnings up and write them out there, where proper logging is available and easy. I don't know exactly how this should be done, but I think it's the cleanest solution. So far down in the program nothing should write to the command line.

nevrome:
I decided to solve this issue now pragmatically with a mixture of 2 and 3. Curious, what you say, @stschiff.

I will use this build in the next couple of days, to find things I might have missed. You could start thinking about how you want to use this for the server. Ready for Review!

nevrome:
ToDo from my own observations:
- `NoLog` should also turn off the progress indicators
- `SimpleLog` should feature a "manual" newline after the progress indicators

stschiff (approved):
Approved upon personal discussion

PR #175, opened 2022-05-31
Json for entities

stschiff:
Mini-PR, just to add JSON capabilities to entities, which comes in handy for `xerxes`, where I'd like to support config files in YAML.

ToJSON and FromJSON Instances are written by hand, as I usually prefer explicit over implicit, even if it comes at the cost of some (minimal) verbosity.

### DIFF
--- a/src/Poseidon/EntitiesList.hs
+++ b/src/Poseidon/EntitiesList.hs
@@ -13,13 +13,14 @@ import           Poseidon.Utils          (PoseidonException (..))
 
 import           Control.Applicative     ((<|>))
 import           Control.Exception       (throwIO)
-import           Data.Aeson              (FromJSON(..), ToJSON(..), withText, Value(..))
-import Data.Aeson.Types (Parser)
+import           Data.Aeson              (FromJSON (..), ToJSON (..),
+                                          Value (..), withText)
+import           Data.Aeson.Types        (Parser)
 import           Data.Char               (isSpace)
 import           Data.Function           ((&))
 import           Data.List               (nub, (\\))
 import           Data.Maybe              (mapMaybe)
-import Data.Text (Text, unpack, pack)
+import           Data.Text               (Text, pack, unpack)
 import qualified Text.Parsec             as P
 import qualified Text.Parsec.String      as P
 


src/Poseidon/EntitiesList.hs
```diff
@@ -13,10 +13,13 @@ import           Poseidon.Utils          (PoseidonException (..))
 
 import           Control.Applicative     ((<|>))
 import           Control.Exception       (throwIO)
+import           Data.Aeson              (FromJSON(..), ToJSON(..), withText, Value(..))
+import Data.Aeson.Types (Parser)
```
nevrome, line 17: We should probably switch to some sort of linter or beautifier to avoid inconsistent formatting.

nevrome, line 17: We should probably switch to some sort of linter or beautifier to avoid inconsistent formatting.

src/Poseidon/EntitiesList.hs
```diff
@@ -78,6 +81,19 @@ instance EntitySpec PoseidonEntity where
         parseInd   = Ind   <$> P.between (P.char '<') (P.char '>') parseName
         parseName  = P.many1 (P.satisfy (\c -> not (isSpace c || c `elem` ",<>*")))
 
+-- turns out that we cannot easily write instances for classes, so need to be explicit for both types
+instance FromJSON PoseidonEntity where parseJSON = withText "PoseidonEntity" aesonParseEntitySpec
+instance FromJSON SignedEntity   where parseJSON = withText "SignedEntity" aesonParseEntitySpec
+instance ToJSON   PoseidonEntity where toJSON e = String (pack $ show e)
+instance ToJSON   SignedEntity   where toJSON e = String (pack $ show e)
+
+aesonParseEntitySpec :: (EntitySpec e) => Text -> Parser e
```
nevrome, line 90: What's the advantage of `Text` here over `String`, if you have to unpack anyway? I read that Text is better than String in many ways, but here it appears a bit random?

nevrome, line 90: What's the advantage of `Text` here over `String`, if you have to unpack anyway? I read that Text is better than String in many ways, but here it seems a bit sporadic?

nevrome (approved):

src/Poseidon/EntitiesList.hs
stschiff: Yeah, I use stylish-haskell from vscode. Have just fixed this layout.

stschiff: Fixed, see above.

src/Poseidon/EntitiesList.hs
stschiff: There is no advantage, but the aeson package requires this at this point. If you look above, I use `withText "PoseidonEntity" aesonParseEntitySpec`, where the last one needs to have the type signature with `Text`.

PR #177, opened 2022-06-02
fixed a bug in fetch

nevrome:
`head` is not safe and `baseDirs` can be an empty list (I think since #172)


stschiff:
I think we need a quick discussion about this. If I understand correctly, the `fetch` doesn't need a base directory anymore. You basically then take `.` implicitly. I think that's dangerous and not intuitive.

stschiff:
OK, I approve this for now.

stschiff (approved):

nevrome:
This suggestion will become obsolete with the changes in #191

PR #187, opened 2022-06-22
switch to GHC 8.10.7

nevrome:
8.10.7 is only 9 months younger than 8.10.3, but it allows us to use the latest HLS version. There would be one younger stackage LTS release ([LTS 19.12 for ghc-9.0.2](https://www.stackage.org/lts-19.12)), but that is only 3 days old and may be unreliable. Also: `table-layout` does not work with GHC 9 yet.


stschiff:
I've just recompiles and it works. OK for me to merge.

stschiff (approved):

nevrome:
Brilliant! Very happy to get rid of 8.10.3, finally.

PR #190, opened 2022-06-23
Add output dir for genoconvert

### LINKED ISSUE #184: genoconvert should create necessary output directory
Using `genoconvert` with `-o` can result in an error "openFile: does not exist (No such file or directory)".

I think we should try to create the output directory if it's given. (I can do that, just a reminder)

stschiff:
Fixes #184

### DIFF
--- a/src/Poseidon/CLI/Fetch.hs
+++ b/src/Poseidon/CLI/Fetch.hs
@@ -69,6 +69,9 @@ runFetch (FetchOptions baseDirs entitiesDirect entitiesFile remoteURL upgrade do
         downloadDir = head baseDirs
         tempDir = downloadDir </> ".trident_download_folder"
     
+    logInfo $ pack $ "Download directory (will be created if missing): " ++ downloadDir
+    liftIO $ createDirectoryIfMissing True downloadDir
+
     -- compile entities
     entitiesFromFile <- liftIO $ mapM readEntitiesFromFile entitiesFile
     let entities = nub $ entitiesDirect ++ concat entitiesFromFile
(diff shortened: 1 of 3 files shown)


nevrome:
For fetch this solution is not sufficient, I fear:
```
$ trident fetch -d huhu -f "*2010_RasmussenNature*"
[Info]    Downloading individual list from remote
[Info]    Downloading package list from remote
trident: huhu: getDirectoryContents:openDirStream: does not exist (No such file or directory)
```
Fetch treats the first input baseDir as the output dir, so if the baseDir list has non-existent entries, then this will fail anyway. Maybe we should give it an output dir argument, just like `forge` and `init` already have it. Then the input baseDirs could become optional for fetch. The out dir could get the default value ".", to really purge the default call: You could then run fetch only with `-f`.

```
trident fetch -f "..."
```

stschiff:
Again too lazy to try it out myself 😩. Sorry. Yes OK, I'll think about it and will get back to it.

stschiff:
Hmm, in principle I like supporting a use case where you want to "just download a package". But we do currently make some cool inferences about Version changes and what needs updating, etc... how should we do this if we have the user specify an outgroup directory? What happens if a package directory already exists in the outgroup place? Are we then conducting a version diff?

I think for now I'd just like to update the behaviour with an empty directory given with `-d` as you demonstrate above currently fails, and then perhaps come back to your suggestion of specifying an outgroup dir.

nevrome:
Naja - the other subcommands have explicit output dirs now. And one could argue, that this is a useful option to have. Of course we should keep `-d` to allow for the check of what is already in your selection.

> What happens if a package directory already exists in the outgroup place?
So if `-o` is specified, but not `-d`, or `-o` is not in `-d`, then we can just overwrite. If a user wants to prevent that, then they need to give `-d`. Or we could also just append `-o` to `-d`. Then the duplicate check would automatically be applied to it as well. In a way that's already what we're doing: We write to the first position of `-d`. With `-o` we could make things just more explicit and give the user more control.

stschiff:
OK, I provided a fix for `fetch`, which preserves the original behaviour, so does aim at the first provided base directory.

nevrome (approved):
Works now! :fireworks:

PR #191, opened 2022-06-24
properly parse base directories and genotype source files

### LINKED ISSUE #182: CLI should properly parse base directories and genotype source files
Currently, with the new direct genotype data interface, the CLI does not complain anymore if no source is input. It would be better to relegate this logic to the type level and introduce something like

```
data SourceSpec = SourceSpecBaseDir FilePath | SourceSpecGenoDirect FilePath
```

and then replace the `OP.many` in `parseBasePaths` with `OP.many1` or `OP.some`.

nevrome:
- fixes #182
- removed short options (-r + -g + -s + -i) for the direct genotype data input

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 docs/_build/
 .stack-work/
+dist-newstyle/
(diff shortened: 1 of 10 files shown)


nevrome:
Would you mind taking another look at this, @stschiff, to complete the review? This introduces breaking changes and would therefore be `v1.1.0.0`. For that we should do a release then, imho, including all the changes from last week.

Issue #183, opened 2022-06-22
From `v0.27`, `--forgeFile` and `--forgeString` can't be appeared together. Can still do it in `v0.26` .


huanglei-artificium:
I would like to merge samples from different packages. For some packages, I want to select tens of individuals, so I use a file; while for others, I only need to select a few of them, and I can use string. I prefer to create a different file/string for different packages.
![image](https://user-images.githubusercontent.com/56875428/174978826-73dbd3fc-0646-48ac-a141-5b25a3b999ef.png)

nevrome:
Solved in #192. Will be part of the next release.

PR #192, opened 2022-06-24
Flexible entity input

stschiff:
This should address #183. I made a few changes to the internal representation of direct Entity input and File input. Now the CLI should already guard against any non-sensical input (for example, using `--downloadAll` and entering entities in `fetch` was previously possible, but it shouldn't, and it isn't with this PR).

Note that despite the internal change, the CLI behaviour hasn't changed (for reasonable input). For example, with `forge`, if the user supplies neither a forgeString nor a forgeFile, it means that everything is forged. In `fetch`, the user _has_ to choose `--downloadAll` or give input through a direct string or file. One might want to debate whether this difference between fetch and forge is OK, but at least this is how it behaved so far. Internally, in both cases we now represent an empty list of strings and files as "forge or fetch all".

### DIFF
--- a/src-executables/Main-trident.hs
+++ b/src-executables/Main-trident.hs
@@ -349,9 +349,9 @@ parseForgeEntitiesDirect = OP.option (OP.eitherReader readSignedEntities) (OP.lo
         \Negative selection is possible by prepending \"-\" to the entity you want to exclude \
         \(e.g. \"*package_1*, -<individual_1>, -group_1\"). \
         \forge will apply excludes and includes in order. If the first entity is negative, then forge \
-        \will assume you want to merge all individuals in the \
-        \packages found in the baseDirs (except the ones explicitly excluded) before the exclude entities are applied. \
-        \An empty forgeString, with no --forgeFile given, will therefore merge all available individuals.")
+        \will assume you want to merge all individuals in the packages found in the baseDirs (except the \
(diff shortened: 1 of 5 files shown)


nevrome:
I fear this branch is thoroughly broken. Both (!) `forge` and `fetch` fail already in the input parsing step with some endless loop that fills the memory and almost grilled my system :smile: Maybe something about accepting empty input? You can reproduce it already with just `trident forge` or `trident fetch`, without any arguments.

Beyond that;
- I think `--downloadAll` is a reasonable "safety measure". Few users want to download everything, so it should be a conscious decision to trigger that. Merging every input dataset should be a more common application, that can very well be made available conveniently without an extra `--forgeAll` flag. Or do you think we should introduce a general `--all` flag for these cases?
- As already mentioned in the chat: If the order of `--forgString`s and `--forgeFile`s arguments is meaningful, then this should be documented.

stschiff:
Oh god. Embarrassing. I relied 100% on the Golden Tests and didn't even bother to just once run the CLI. Lesson learned. Will fix ASAP

stschiff:
OK, this should be fixed now.

nevrome:
Nice! I think it works now exactly as expected. Very neat! I added a sentence to the documentation as suggested. We have to remember to update the documentation on the website when all of these PRs are merged.

nevrome (approved):

PR #194, opened 2022-07-13
Getting logging work in pipes

stschiff:
I have now solved this, at least conceptually, by basically changing the Monad Stack from the previous `LoggerT Message IO` to a `LoggerT Message (SafeT IO)`. This relatively simple, by mostly using `MonadSafe` as a constraint in the CLI functions.


stschiff:
OK, this compiles now and tests succeed. But lot of todos still:
1) Getting proper logging in the genotype pipe, without the current forwarding of message lists
2) Testing for performance or memory leaks
3) Fixing loggin also in the Server

PR #195, opened 2022-07-13
stylish-haskell

nevrome:
As discussed: We should either abandon auto-formatting or fully embrace it. It's not helpful if you do it and I don't, @stschiff.

Here I set up a default `.stylish-haskell.yaml` (please check and adjust to your liking) and run it once. If merged, this should immediately make #194 more readable.

I set up a git hook to remind myself to run it, but that's probably not very reliable. So maybe we should make it part of the GitHub action. But I'm not 100% sure how. I think `stylish-haskell` has no "check if anything WOULD change"-mode, right? We could run it, change the files in place and check if that differs from the committed state :thinking:


stschiff:
Wow, cool, I'll take a look.

stschiff (approved):
OK, obviously can't review all these file. But looks good. Let's merge this in!

stschiff:
OK, don't fully get the close, but I gather stylish-haskell has been moved in our routine checks anyway? Good.

nevrome:
I think this had sth. to do with the other open PRs and possible merge conflicts. I added the feature somewhere else.

PR #196, opened 2022-07-19
Adapted Logging

stschiff:
My motivation for this PR is two-fold:
1) I wanted to clean up this hack of collecting the SNP-merging messages from inside the `SafeT` Pipe in a list before logging them. I think it obfuscates the code. I think if a function needs to log something, it simply needs to be able to do so, no matter where in the function call stack we are. Perhaps that's a bit up to taste... but given the emerged complexity of this project I jump at opportunities to simplify things.

2) I found it unsatisfactory that we already use a `ReaderT`-like system to pass Logging Actions, but are unable to pull this LogAction when entering the `SafeT pipe`, making this hack of passing `LogMode` from the top to the very bottom. 

The key change is a change of `PoseidonLogIO`. Before, we had used `LoggerT`, which is a wrapper around this:

```
ReaderT (LogAction (LoggerT Message IO) Message) IO
```

which looks scary, and it is. The tricky bit here is that because of this inner `LoggerT` type inside of `LogAction`, there is no way we can get out the action to perform logging inside IO. For example, we might want to say `logAction <- ask` inside of `PoseidonLogIO`, which is possible because of the `MonadReader` instance of `LoggerT`. This would be cool, because then we don't have to pass `LogMode` from the top, but can use `logAction` where we need it. 

_BUT_: The thing returned from `ask` is actually a `LogAction (LoggerT IO Message) Message IO`, which is extremely annoying and essentially forbids me to use this inside of `MonadIO`. 

I found a simple solution, which simply doesn't use `LoggerT`, but the much simpler:
```
type PoseidonLogIO = ReaderT (LogAction IO Message) IO
```

Note that the difference is the internal monad in `LogAction`, which is now `IO`, not `LoggerT Message IO`.

(Note that we actually define `type LogEnv = LogAction IO Message` in the code to make this even more readable)

Now, this works, but unfortunately it doesn't work with the existing `logDebug`, `logError` etc. functions from CoLog. Easy to fix, we just define our own, right in `Utils`... just 10 lines and we're done. So that's the "ugly" price to pay, but at least it's only ugly inside of `Utils`. All client code actually simplifies. First, I was able to sneak in `pack`, so we can now log Strings, getting rid of like a million or so `pack` calls. Second, as promised above, we can now use `ask` to retrieve the logging environment and pass it to client functions like `getJointGenotypeData` and further down. So I was able to remove `LogMode` from `ForgeOptions`.

I tried a quick forge and did not see any performance or memory leak.

### DIFF
--- a/src/Poseidon/Utils.hs
+++ b/src/Poseidon/Utils.hs
@@ -1,49 +1,65 @@
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE OverloadedStrings, FlexibleInstances, MultiParamTypeClasses #-}
 
 module Poseidon.Utils (
     PoseidonException (..),
     renderPoseidonException,
     usePoseidonLogger,
     PoseidonLogIO,
-    LogMode (..)
+    LogMode (..),
+    logWarning,
+    logInfo,
+    logDebug,
+    logError,
+    LogEnv,
+    noLog
 ) where
 
-import           Colog                  (LoggerT, Message, usingLoggerT, LogAction (..), cmapM, cfilter,
-                                         logTextStderr, showSeverity, msgSeverity, msgText, Severity (..))
-import           Control.Exception      (Exception, try, IOException)
+import           Colog                  (LogAction (..), Message,
+                                         Severity (..), cfilter, cmapM,
+                                         logTextStderr, msgSeverity, msgText,
+                                         showSeverity, Msg(..), HasLog(..))
+import           Control.Exception      (Exception, IOException, try)
 import           Control.Monad          (when)
-import           Data.Yaml              (ParseException)
+import           Control.Monad.IO.Class (liftIO)
+import           Control.Monad.Reader   (ReaderT, runReaderT, asks)
 import           Data.Text              (Text, pack)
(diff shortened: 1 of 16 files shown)


src/Poseidon/Utils.hs
```diff
@@ -59,40 +76,65 @@ compileLogMsg severity time cursorCheck = cmapM prepareMessage logTextStderr
...
+    | PoseidonFStatsFormatException String
+    | PoseidonBibTeXException FilePath String
+    | PoseidonPoseidonEntityParsingException String
+    | PoseidonForgeEntitiesException String
+    | PoseidonEmptyForgeException
+    | PoseidonNewPackageConstructionException String
+    | PoseidonRemoteJSONParsingException String
+    | PoseidonGenericException String
+    | PoseidonEmptyOutPacNameException
+    | PoseidonUnequalBaseDirException FilePath FilePath FilePath
```
nevrome, line 137: Wait - why was that necessary? I need these explanations...

nevrome (approved):

nevrome:
I want to add that I also experimented with `ask` and just couldn't find a way to do it. So thanks for showing and explaining!

stschiff:
Yes sorry again for not reviewing the other PRs first.

PR #197, opened 2022-07-25
outsourced optparse-applicative parsers to an own library module

nevrome:
This is helpful for xerxes and other derived tools/libraries


stschiff:
Great, thank you!

PR #198, opened 2022-07-26
Simplify progress bars

nevrome:
An experiment to switch to simple, sequential progress indication

### DIFF
--- a/src-executables/Main-trident.hs
+++ b/src-executables/Main-trident.hs
@@ -100,8 +100,7 @@ renderVersion :: String
 renderVersion =
     "trident v" ++ showVersion version ++ " for poseidon v" ++
     intercalate ", v" (map showPoseidonVersion validPoseidonVersions) ++ "\n" ++
-    "https://poseidon-framework.github.io" ++ "\n" ++
-    ")<(({°> ~ ────E ~ <°}))>("
+    "https://poseidon-framework.github.io"
 
 subcommandParser :: OP.Parser Subcommand
 subcommandParser = OP.subparser (
--- a/src/Poseidon/CLI/Fetch.hs
+++ b/src/Poseidon/CLI/Fetch.hs
@@ -13,7 +13,7 @@ import           Poseidon.Package        (PackageReadOptions (..),
                                           readPoseidonPackageCollection)
 import           Poseidon.SecondaryTypes (IndividualInfo (..), PackageInfo (..))
 import           Poseidon.Utils          (PoseidonException (..), PoseidonLogIO,
-                                          logInfo, logWarning)
+                                          logInfo, logWarning, LogEnv, logWithEnv)
 
 import           Codec.Archive.Zip       (ZipOption (..),
                                           extractFilesFromArchive, toArchive)
@@ -22,6 +22,7 @@ import           Conduit                 (ResourceT, await, runResourceT,
 import           Control.Exception       (throwIO)
 import           Control.Monad           (forM_, unless, when)
 import           Control.Monad.IO.Class  (liftIO)
+import           Control.Monad.Reader    (ask)
 import           Data.Aeson              (eitherDecode')
 import qualified Data.ByteString         as B
 import           Data.ByteString.Char8   as B8 (unpack)
@@ -32,11 +33,9 @@ import           Network.HTTP.Conduit    (http, newManager, parseRequest,
                                           responseBody, responseHeaders,
                                           simpleHttp, tlsManagerSettings)
 import           Network.HTTP.Types      (hContentLength)
-import           System.Console.ANSI     (hClearLine, hSetCursorColumn)
 import           System.Directory        (createDirectoryIfMissing,
                                           removeDirectory, removeFile)
 import           System.FilePath         ((</>))
-import           System.IO               (hFlush, hPutStr, stderr)
 
 data FetchOptions = FetchOptions
     { _jaBaseDirs  :: [FilePath]
@@ -162,9 +161,10 @@ printV (Just x) = showVersion x
 
 downloadAndUnzipPackage :: FilePath -> FilePath -> String -> String -> PoseidonLogIO ()
 downloadAndUnzipPackage baseDir tempDir remote pacName = do
-    logInfo $ padString 40 pacName
+    logInfo $ padString 40 pacName ++ "now downloading"
+    logEnv <- ask
     liftIO $ do
-        downloadPackage tempDir remote pacName
+        downloadPackage logEnv tempDir remote pacName
         unzipPackage (tempDir </> pacName) (baseDir </> pacName)
         removeFile (tempDir </> pacName)
 
@@ -174,8 +174,8 @@ unzipPackage zip_ outDir = do
     let archive = toArchive archiveBS
     extractFilesFromArchive [OptRecursive, OptDestination outDir] archive
 
-downloadPackage :: FilePath -> String -> String -> IO ()
-downloadPackage pathToRepo remote pacName = do
+downloadPackage :: LogEnv -> FilePath -> String -> String -> IO ()
+downloadPackage logEnv pathToRepo remote pacName = do
     downloadManager <- newManager tlsManagerSettings
     packageRequest <- parseRequest (remote ++ "/zip_file/" ++ pacName)
     runResourceT $ do
@@ -183,8 +183,9 @@ downloadPackage pathToRepo remote pacName = do
         let Just fileSize = lookup hContentLength (responseHeaders response)
         let fileSizeKB = (read $ B8.unpack fileSize) :: Int
         let fileSizeMB = roundTo 1 (fromIntegral fileSizeKB / 1000.0 / 1000.0)
+        logWithEnv logEnv $ logInfo $ "Package size: " ++ show (roundTo 1 fileSizeMB) ++ "MB"
         sealConduitT (responseBody response) $$+-
-            printDownloadProgress fileSizeMB .|
+            printDownloadProgress logEnv fileSizeMB .|
             sinkFile (pathToRepo </> pacName)
     return ()
 
@@ -194,8 +195,8 @@ padString n s
     | length s < n  = s ++ replicate (n - length s) ' '
     | otherwise     = s
 
-printDownloadProgress :: Double -> ConduitT B.ByteString B.ByteString (ResourceT IO) ()
-printDownloadProgress fileSizeMB = loop 0 0
+printDownloadProgress :: LogEnv -> Double -> ConduitT B.ByteString B.ByteString (ResourceT IO) ()
+printDownloadProgress logEnv fileSizeMB = loop 0 0
     where
         loop loadedB loadedMB = do
             x <- await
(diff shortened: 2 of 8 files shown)


src-executables/Main-trident.hs
```diff
@@ -100,8 +100,7 @@ renderVersion :: String
 renderVersion =
     "trident v" ++ showVersion version ++ " for poseidon v" ++
     intercalate ", v" (map showPoseidonVersion validPoseidonVersions) ++ "\n" ++
-    "https://poseidon-framework.github.io" ++ "\n" ++
-    ")<(({°> ~ ────E ~ <°}))>("
+    "https://poseidon-framework.github.io"
```
stschiff, line 103: 😂 I must have put you off with my questions about this ASCII art... sorry.

src/Poseidon/CLI/Fetch.hs
```diff
@@ -174,17 +175,18 @@ unzipPackage zip_ outDir = do
     let archive = toArchive archiveBS
     extractFilesFromArchive [OptRecursive, OptDestination outDir] archive
 
-downloadPackage :: FilePath -> String -> String -> IO ()
-downloadPackage pathToRepo remote pacName = do
+downloadPackage :: LogEnv -> FilePath -> String -> String -> IO ()
```
stschiff, line 178: I think this function could still run in `PoseidonLogIO`, can't it? only in `printDownloadProgress` you need to switch to the manual handover of `logEnv`?

src/Poseidon/GenotypeData.hs
```diff
@@ -256,18 +259,30 @@ recodeAlleles consensusSnpEntry snpEntry genoLine = do
...
-            liftIO $ hPutStr stderr ("> " ++ show n ++ " ")
-            liftIO $ hFlush stderr
+        currentTime <- liftIO getCurrentTime
+        logWithEnv logEnv $ logProgress n (diffUTCTime currentTime startTime)
         liftIO $ modifyIORef counterRef (+1)
         yield val
+    where
+        logProgress :: Int -> NominalDiffTime -> PoseidonLogIO ()
+        logProgress c t
+            | c >=  1000000 &&                c `rem` 1000000 == 0 = logInfo $ "SNPs: " ++ pad (show c) ++ " - " ++ prettyTime t
```
stschiff, line 274: Very cool idea to stack it with order of magnitude! I will have to try that out myself, but code-wise definitely looks very cool!

src/Poseidon/GenotypeData.hs
```diff
@@ -256,18 +259,30 @@ recodeAlleles consensusSnpEntry snpEntry genoLine = do
...
+            | c >=  100000  && c <= 500000 && c `rem` 100000  == 0 = logInfo $ "SNPs: " ++ pad (show c) ++ " - " ++ prettyTime t
+            | c >=  10000   && c <= 50000  && c `rem` 10000   == 0 = logInfo $ "SNPs: " ++ pad (show c) ++ " - " ++ prettyTime t
+            |                  c <= 5000   && c `rem` 1000    == 0 = logInfo $ "SNPs: " ++ pad (show c) ++ " - " ++ prettyTime t
+            | otherwise = return ()
+        pad :: String -> String
+        pad s
+            | length s < 9  = replicate (9 - length s) ' ' ++ s
+            | otherwise     = s
+        prettyTime :: NominalDiffTime -> String
+        prettyTime = formatTime defaultTimeLocale "%H:%M:%S" . posixSecondsToUTCTime
```
stschiff, line 284: Isn't there a simpler way to do this time-wise logging? I would think that, if we want a time-stamp in our logging, we should do that higher up in our log-function? Or do you think it's important to have time-stamps only here in the SNP progress, but not in all other logging places? I don't really see why this distinction is necessary... either time-stamps are important for debugging logging-information, or not, but why would they be interesting in some cases but not others?

src/Poseidon/Package.hs
```diff
@@ -299,16 +297,9 @@ readPoseidonPackageCollection opts dirs = do
...
-            hClearLine stderr
-            hSetCursorColumn stderr 0
-            hPutStr stderr $ "> " ++ show numberPackage ++ " "
-            hFlush stderr
-        try . readPoseidonPackage opts $ path
-    tryDecodePoseidonPackage True (numberPackage, path) = do
-        liftIO $ hPutStrLn stderr $ "> " ++ show numberPackage ++ ": " ++ path
+    tryDecodePoseidonPackage :: (Integer, FilePath) -> PoseidonLogIO (Either PoseidonException PoseidonPackage)
+    tryDecodePoseidonPackage (numberPackage, path) = do
+        logDebug $ "Package " ++ show numberPackage ++ ": " ++ path
```
stschiff, line 302: what happened here? Remove without replacement?

stschiff (commented):
Excellent. Thanks! Code-wise looks good, see some comments/questions. I will try it out myself now.

src-executables/Main-trident.hs
nevrome: Stephen also didn't understand it, and asked if something is wrong with the terminal output. So I guess then it's just confusing and should be scrapped. I'm not tooo sad about it :slightly_smiling_face:

stschiff: Haha... OK

src/Poseidon/GenotypeData.hs
nevrome: I read that one should avoid slowing the progress indication down. For psychological reasons. So I'm not sure if this is really the best solution. Could we easily (and quickly) determine the number to SNPs to be merged in a given forge operation? Then we could do a linearly scaled progress update. :thinking:

stschiff: Hmm... I suppose quickly counting the lines of the bim file would do the job. I just tried with `wc -l` on the command line and with 1240K data it goes pretty much instantaneous. I think using my bim-reader from `sequence-formats` will be slower, since it also parses it into Haskell structures. Perhaps there is a quick line-counting routine somewhere in the Haskell world? The turtle library might have a function for that... but again, whether it's as fast as `wc -l` I don't know.

stschiff:
OK, I have played with this. Works like a charm, but I now have clearer opinions:

1) I think the non-linear printing of SNP conversion is indeed a bit confusing. I think I would try and find a linear logging with the assumption that 99% of use cases will have around 1,2 million SNPs... so I think printing a line every 10,000 SNPs seems acceptable to me. If people forge complete-genome data with 60M SNPs that means 6000 log-lines, which is the maximum and still seems not too terrible to me.

2) I now understand that the time-stamps after the SNP count denotes time passed since the start of the download, not timestamp of the log, which seems less redundant than I thought above. We can keep that, I guess. I think if I had implemented that, though, I wouldn't have put this in, since the user in principle has a way to deal with this using `--logMode ServerLog` (which I think we should rename or redesign to make it easier to print time stamps).

`ServerLog` looks quite beautiful in this case:
```
[Info]    [13:14:53] Searching POSEIDON.yml files... 
[Info]    [13:14:53] 1 found
[Info]    [13:14:53] Checking Poseidon versions... 
[Info]    [13:14:53] Initializing packages... 
[Info]    [13:14:53] Packages loaded: 1
[Info]    [13:14:53] Unpackaged genotype data files loaded: 0
[Info]    [13:14:53] Converting genotype data in 2012_PattersonGenetics to format EIGENSTRAT:
[Info]    [13:14:53] Processing SNPs...
[Info]    [13:14:53] SNPs:         0 - 00:00:00
[Info]    [13:14:53] SNPs:      1000 - 00:00:00
[Info]    [13:14:53] SNPs:      2000 - 00:00:00
[Info]    [13:14:54] SNPs:      3000 - 00:00:00
[Info]    [13:14:54] SNPs:      4000 - 00:00:00
[Info]    [13:14:54] SNPs:      5000 - 00:00:00
[Info]    [13:14:54] SNPs:     10000 - 00:00:00
[Info]    [13:14:55] SNPs:     20000 - 00:00:01
[Info]    [13:14:56] SNPs:     30000 - 00:00:02
[Info]    [13:14:56] SNPs:     40000 - 00:00:03
[Info]    [13:14:57] SNPs:     50000 - 00:00:03
[Info]    [13:15:01] SNPs:    100000 - 00:00:07
[Info]    [13:15:08] SNPs:    200000 - 00:00:14
[Info]    [13:15:16] SNPs:    300000 - 00:00:22
...
```

So I think that would make the time stamps redundant (you know, keeping this simpler)... but I don't mind keeping it now, as you have already implemented it.

stschiff:
So I think I approve this pending our discussion on the non-linear progress-logging. Happy to be convinced if you prefer the non-linear setup...

src/Poseidon/GenotypeData.hs
nevrome: To discuss this at one place:
> I think the non-linear printing of SNP conversion is indeed a bit confusing. I think I would try and find a linear logging with the assumption that 99% of use cases will have around 1,2 million SNPs... so I think printing a line every 10,000 SNPs seems acceptable to me. If people forge complete-genome data with 60M SNPs that means 6000 log-lines, which is the maximum and still seems not too terrible to me.

6000 lines is ridiculous, but how often will somebody merge that many SNPs? Maybe the simplest solution is really the best. 10000 doesn't sound too bad.

src/Poseidon/GenotypeData.hs
nevrome: > I now understand that the time-stamps after the SNP count denotes time passed since the start of the download, not timestamp of the log, which seems less redundant than I thought above. We can keep that, I guess. I think if I had implemented that, though, I wouldn't have put this in, since the user in principle has a way to deal with this using --logMode ServerLog (which I think we should rename or redesign to make it easier to print time stamps).

I thought it may be useful for the user to quickly estimate how long a given process will still take. To do that "right", we would need an estimate of the number of total SNPs to be merged (as discussed above). Then we can give a proper ETA.

src/Poseidon/Package.hs
nevrome: Simplification of the progress indicator. It just got a lot simpler and can not be turned on/off any more.

src/Poseidon/GenotypeData.hs
nevrome: I decided to implement this now. It's verbose, but it is 100% clear and reliable.

src/Poseidon/GenotypeData.hs
nevrome: I implemented a human-optimized time display now:
```
[Info]    SNPs:    420000    55s
[Info]    SNPs:    430000    56s
[Info]    SNPs:    440000    58s
[Info]    SNPs:    450000    59s
[Info]    SNPs:    460000    1m 0s
[Info]    SNPs:    470000    1m 2s
[Info]    SNPs:    480000    1m 3s
[Info]    SNPs:    490000    1m 4s
```
I think that adds enough value to keep it around independent of the timestamp one can get through the logging infrastructure.

src/Poseidon/CLI/Fetch.hs
nevrome: I refactored that bit now.

nevrome:
I think I'm happy with it now. Can you take another quick look?

stschiff:
OK, looks very good to me. There is really only one change that I need to now think about on the server. You previously had a nice little progress indicator for the packages as they're being loaded, which now got removed. There is still a `logDebug` though. Since on the server the loading takes some time, since I'm usually checking the checksums, I'm thinking whether I can live with the Debug output. I'll get back to you.

stschiff:
OK, I think I will then simply switch to `VerboseLog` on the server, in order to see the package loading. I think that's fine then. I cannot implement this within this PR, because I only switched to colog in #200. So good to accept!

nevrome:
Yes - right. I decided to move this to the `debug` level, because it's 150 lines for our public dataset and quite often not really relevant. Sorry for breaking your workflow here.

stschiff:
All good.

PR #200, opened 2022-08-11
Server updates

stschiff:
I have updated the server in multiple ways:
1) It now uses Co-Log for logging
2) A new option `-c` now makes it ignore checksums, which is useful for a fast start of the server if need be
3) Zip files are now stored in a separate folder, to keep the (git-backed) repository itself clean
4) There is a new API named `/compatibility/<version>` which accepts a client version (from trident) and returns a JSON tuple of Haskell-type (Bool, Maybe String). The first element is simply a Boolean saying if the client version is compatible with the server or not, the second is an optional Warning message the server can return to the client. 

Currently I don't make much use of the new API, but I would like to establish it in time to be able to later get away with a) breaking API changes and b) even more severe - a move of the server to a new URL. The latter will likely happen at some point, as I consider the current Cloud-Server to be temporary and the URL poor to remember. I'd like to switch to something like `www.poseidon-adna.de/server` or so. Anyway, this new API gives us a way of implementing a message-system within `trident` to be able to give early warnings to Clients if the Server intends to introduce breaking changes or a move.

I don't think a Code review is required here, since this PR affects only the server program so far. But of course, if there are objections to the new API, bring it on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- V 1.1.3.0: Added new features to the server, updated logging, and new API for compatibility checks
 - V 1.1.2.0: Replaced progress indicators with simple, sequential log messages for the download in fetch and the SNP-wise operations in forge and genoconvert
 - V 1.1.1.3: Tiny change to make the documentation of --snpSet more clear
 - V 1.1.1.2: Outsourced optparse-applicative parsers to an own library module. This is helpful for xerxes and other derived tools/libraries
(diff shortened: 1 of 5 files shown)


stschiff:
I just switched to `VerboseLog` in order to still see the packages as they're being loaded. Due to a recent change in #198, the progress indication upon package loading was removed, but on the server loading times are long due to checksum checks, so I would like to see those.

stschiff:
And in response to switching to VerboseLog I have now also commented out two types of Debug messages from the SNP loading, which indicated that some SNPs are monomorphic or have missing data, which is actually quite normal and doesn't need a Debug message.

nevrome:
These changes are all fine to me :+1:
The `/compatibility/` interface is brilliant! Makes updates much more safe.

nevrome:
I merged #198 and gave it the version number 1.1.2.0. I suggest you iterate to 1.1.3.0 for the changes here. I already wrote a release changelog for 1.1.3.0. Feel free to add something to it before creating the tag.

PR #201, opened 2022-08-18
Catch missing bibentries if bibtex file is missing

### LINKED ISSUE #199: bib-file consistency check in validate is leaky
I'm looking at [the code lines](https://github.com/poseidon-framework/poseidon-hs/blob/d9249712aa4ef2332928c2eb7cda6ebb864f1e4d/src/Poseidon/Package.hs#L338) which call `checkJannoBibConsistency`, and I think it's a problem that they are only called when a bib-file is present. I think there are cases where - for example - there are bib-keys in the Janno file, but no bibtex-file present, which should obviously be flagged.

stschiff:
Fixes #199. The key changes in this PR are in `src/Poseidon/Package.hs` and `test/Poseidon/PackageSpec.hs`. 

**Everything else is just linter and style fixes** (I finally git HLS working and followed some linter advices... sorry if it clutters this PR)

### DIFF
--- a/src/Poseidon/CLI/Forge.hs
+++ b/src/Poseidon/CLI/Forge.hs
@@ -96,7 +96,7 @@ runForge (
 
     entities <- case entitiesUser of
         [] -> do
-            logInfo $ "No requested entities. Implicitly forging all packages."
+            logInfo "No requested entities. Implicitly forging all packages."
             return $ map (Include . Pac . posPacTitle) allPackages
         (Include _:_) -> do
             return entitiesUser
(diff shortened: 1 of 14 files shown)


src/Poseidon/CLI/Forge.hs
```diff
@@ -191,7 +191,7 @@ runForge (
                     P.tee outConsumer
             let startAcc = liftIO $ VUM.replicate (length newEigenstratIndEntries) 0
             P.foldM sumNonMissingSNPs startAcc return forgePipe
-        ) (\e -> throwIO $ PoseidonGenotypeExceptionForward e)
+        ) (throwIO . PoseidonGenotypeExceptionForward)
```
nevrome, line 194: Haha - ok - HLS already suggested this change to me, but I did not follow it, because I find the explicit lambda function more readable. Same for multiple other changes here :sweat_smile:

nevrome (commented):
I tested it quickly and it seems to behave as expected :+1:

nevrome (approved):

src/Poseidon/CLI/Forge.hs
stschiff: Yeah, I didn't follow all HLS linter suggestions, but this one I liked. Definitely up to taste, and I'm happy to skip these in the future if you prefer the lambdas. Good to know.

PR #205, opened 2022-11-21
Poseidon v2.6

nevrome:
This PR should eventually include all changes for the switch to Poseidon v2.6.0 (https://github.com/poseidon-framework/poseidon2-schema/pull/51)


nevrome:
I think that covers everything. Please check, @stschiff.

I'm not super happy about the solution to make the contributor field optional. I used `Maybe [ContributorSpec]` in `PoseidonYamlStruct`, because aeson makes it easy to handle `Maybe`. But an empty list represents this case just as well, so I left it at `[ContributorSpec]` in `PoseidonPackage`. Then we also don't have to change any code touching this field. Still this solution is ugly, because it requires two conversions:

```haskell
let con = case maybeCon of
        Nothing -> []
        Just xs -> xs

let maybeCon = case con of
        [] -> Nothing
        xs -> Just xs
```

Do you see an option to read a missing value directly as a list?

src/Poseidon/Package.hs
```diff
@@ -83,7 +83,7 @@ data PoseidonYamlStruct = PoseidonYamlStruct
     { _posYamlPoseidonVersion :: Version
     , _posYamlTitle           :: String
     , _posYamlDescription     :: Maybe String
-    , _posYamlContributor     :: [ContributorSpec]
+    , _posYamlContributor     :: Maybe [ContributorSpec]
```
stschiff, line 86: There is generally a redundancy in a type of the sort `Maybe [Type]`, so of a maybe of a list, since an empty list already signals a "Nothing". I think we can simplify this code. I'll take a stab at that. That there is now a redundancy can be seen in the fact that the two states of this value `Just []` and `Nothing` are semantically equivalent, but differently expressed, which is not ideal.

src/Poseidon/SecondaryTypes.hs
```diff
@@ -123,4 +131,60 @@ oneContributorSpecParser :: P.Parser ContributorSpec
...
+instance FromJSON ORCID where
+    parseJSON (String s) = case P.runParser parseORCID () "" (unpack s) of
+        Left err -> fail $ show err
+        Right x  -> pure x
+    parseJSON _          = mzero
+
+instance ToJSON ORCID where
+    toJSON x = String $ pack $ renderORCID x
+
+parseORCID :: P.Parser ORCID
```
stschiff, line 154: Very nice!

stschiff (commented):
Very nice. Just one comment about the redundancy of `Maybe [ContributorSpec]` which I'd like to improve.

stschiff:
> I think that covers everything. Please check, @stschiff.
> 
> I'm not super happy about the solution to make the contributor field optional. I used `Maybe [ContributorSpec]` in `PoseidonYamlStruct`, because aeson makes it easy to handle `Maybe`. But an empty list represents this case just as well, so I left it at `[ContributorSpec]` in `PoseidonPackage`. Then we also don't have to change any code touching this field. Still this solution is ugly, because it requires two conversions:
> 
> ```haskell
> let con = case maybeCon of
>         Nothing -> []
>         Just xs -> xs
> 
> let maybeCon = case con of
>         [] -> Nothing
>         xs -> Just xs
> ```
> 
> Do you see an option to read a missing value directly as an empty list?

Oh sorry, I just saw this above, so addressing directly my point. Yes, I think aeson can read an empty list just like that. I can make some tests, but I'm confident that should just work out of the box. I think Lists implement the `MonadZero` Typeclass, like `Maybe`s, and I guess Aeson is generic for this type class. Not sure though.

nevrome:
I looked at this again and found a way to make things slightly more elegant in [4e295ec](https://github.com/poseidon-framework/poseidon-hs/pull/205/commits/4e295ec2a68d4a4e360d05d3353559376b9370d5).

This means though, that we get `contributor: []` for the minimal template of the .yml file, which is less elegant than having it completely gone. But I can live with it.

stschiff:
I think I've fixed the contributor output. If it's empty, it's not output.

nevrome:
Excellent!

Issue #202, opened 2022-08-27
Automatic validation of dates?


sc13-bioinf:
I just noticed I messed up the dates for 2021_SaupeCurrBiol. Instead of BCE_CE (can we move to the secular terms please) I used calBP. Clearly a date in the future should not exist.

nevrome:
True - we can check for dates > 2022. Of course it's still possible to enter wrong dates, but this particular case we can safely cover

PR #206, opened 2022-11-29
added a date check as suggested in #202

nevrome:
Just a simple check we can easily do. As suggested in #202.

### DIFF
--- a/src/Poseidon/Janno.hs
+++ b/src/Poseidon/Janno.hs
@@ -8,6 +8,7 @@ module Poseidon.Janno (
     JannoSex (..),
     JannoList (..),
     Sex (..),
+    BCADAge (..),
     Latitude (..),
     Longitude (..),
     JannoDateType (..),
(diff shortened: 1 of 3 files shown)


src/Poseidon/Janno.hs
```diff
@@ -97,6 +98,30 @@ instance Show JannoSex where
...
+
+instance ToJSON BCADAge where
+    toEncoding = genericToEncoding defaultOptions
+
+instance FromJSON BCADAge
+
+instance Csv.FromField BCADAge where
+    parseField x = do
+        val <- Csv.parseField x
+        if val >= 2022 -- the current year
```
stschiff, line 114: Haha, now we have a once-per-year maintenance task to do here :-) Better get the current year programmatically, right? I'll quickly look into it.

stschiff (commented):
Nice! Definitely useful. Let's see whether we can get rid of the hard-coded 2022...

stschiff:
https://stackoverflow.com/questions/52490035/how-to-get-date-in-haskell

src/Poseidon/Janno.hs
nevrome: Ah - right - that was stupid on my end :facepalm:

nevrome:
Alright - thanks!

Issue #109, opened 2021-06-18
Replace cassava with a handwritten .janno parser


nevrome:
In preparation for the CAA I [dived a bit](https://gist.github.com/nevrome/20186b9f02bbea3aa3a383cbdd3e57e1) into parsec and I think now, that we can achieve more precise parsing and - most importantly - better error reporting with a handwritten solution. Either with parsec or megaparsec.

This is not important for now, but a long-to-mid range goal

stschiff:
Oh yes, parsec is great. Unfortunately for all the stream-processing happening in `sequence-formats` I had to use attoparsec because it's the one supported by the pipes framework (it sacrifices error message clarity for speed and is the one the pipes people have chosen to use).

But for anything happening inside of `poseidon-hs`, like Janno and Bibtex parsing, parsec would be good. What place did you have particularly in mind for this?

nevrome:
I see - good to understand the reasoning behind your use of attoparsec!

As the title says I would like to replace cassava for janno parsing. Cassava fails for the whole row and gives pretty imprecise error messages. With parsec we could return better error messages and get rid of a dependency. 

I heard that megaparsec is particularly good with useful error messages, but I have no experiences with it yet. I also heard that parsec is a bit dead regarding new development, but maybe it's just already perfect. How bad can it be, if even pandoc relies on it?

stschiff:
Ah sorry, didn't see the title. YES, I emphatically agree. I didn't dare to say something, because you implemented a pretty nice and working scheme with cassava, but if I would have done this I'd have used parsec for exactly that reason of having a bit more control over the error messages and stuff. So I definitely support this!

nevrome:
Another thing one could theoretically solve with a custom parser is handling non-standard columns. Right now `forge` throws away all columns that are not defined in the Poseidon schema. A better implementation would probably be able to merge supplementary columns as well, setting them to `n/a` if they are missing in an input package.

stschiff:
Good point!

nevrome:
Cassava turned out to be more flexible and powerful than we thought back then. We now have helpful error messages for janno parsing on `--logMode VerboseLog`, e.g.

```
[Debug]   [12:55:59] Can't read sample in ./2010_RasmussenNature.janno in line 2: parse error (Failed reading: conversion error: Latitude "1169.2438" not between -90 and 90)
```

And we even have a mechanism to handle arbitrary additional .janno columns in forge since v1.1.6.0.

[...]

PR #207, opened 2022-12-03
Additional janno columns

nevrome:
This PR implements [a mechanism](https://stackoverflow.com/questions/74598030/haskell-cassava-data-csv-carry-along-additional-columns) to carry along additional, unspecified variables in .janno files. This is a long-desired (see #109) key feature to improve the flexibility of .janno files even more.

For now this is an early draft, which needs thorough testing to make sure this functionality does not break any workflow and works exactly as expected.


nevrome:
I think this works now. But it's a pretty significant change, so I'm a bit insecure if I might have forgotten about something in the bigger scheme of things.  For example the way I ignored this new functionality for JSON conversion.

Could you please review, @stschiff?

stschiff:
OK, sorry for being late on this, but not forgotten.

src/Poseidon/Janno.hs
```diff
@@ -663,64 +739,37 @@ instance Csv.ToNamedRecord JannoRow where
...
-
 jannoHeaderString :: [String]
 jannoHeaderString = map Bchs.unpack jannoHeader
 
+makeHeaderWithAdditionalColumns :: [JannoRow] -> Csv.Header
+makeHeaderWithAdditionalColumns ms =
+    V.fromList $ jannoHeader ++ sort (HM.keys (HM.unions (map (getCsvNR . jAdditionalColumns) ms)))
+
+concatJannos :: [[JannoRow]] -> [JannoRow]
+concatJannos = foldl' combineTwoJannos []
```
stschiff, line 756: Nice!

stschiff (approved):
Amazing job. Thanks! I've gone through the code changes and looks all good to me. I have so little capacity right now that I'm afraid I can't test it, but I think we should just go for it. You've added the tests and if there is a problem we'll fix it in a patch. So good to merge!

nevrome:
Alright - thanks! I'll merge but do some more testing and additional, minor changes in #209 before a new release.

PR #212, opened 2023-01-17
switch from String to Text in the representation of .janno files

### LINKED ISSUE #213: Encoding issue when writing .janno files
The C14 lab number `MAMS-47224 ` becomes `MAMS-47224Â `, or more precisely `"MAMS-47224\160"`  becomes `"MAMS-47224\195\130\194\160"`. Maybe there are more of these encoding issues.

This issue seems to stem from `encodeByNameWith` from `Data.Csv`, which is called in `writeJannoFile`.

nevrome:
This PR replaces `String` in .janno files with `Text`. This has various consequences:
- At some points janno data interacts with other data structures (some inherited from other libraries). There we need a lot of ugly plumbing with `T.pack` and `T.unpack`.
- This is also true for every point where we print janno information to the command line.

When I went through the code for `list` and `summarize` I did some minor refactoring. In `list` I changed the error management for the `-j` option. In `summarize` I simplified the frequency reporting code. The output has now some ugly quotes, but the code is more easy to read. A fair tradeoff, imho.

Finally I also wanted to address the issue that triggered this entire PR: #213. I introduced the new function `cleanInput` in `Janno.hs` which trims white spaces before the actual CSV parsing starts. This is an expensive operation, as I decided to decode the raw input `ByteString`s to `Text`, manipulate them and then encode them again, before they are yet again parsed depending on the output type.
Removing UTF-8 (!) white spaces in `ByteString` was not reliable as some of them are sometimes (?) encoded by two characters (e.g. [`\194\160` for the No-Break Space `\160`](https://stackoverflow.com/questions/13992934/how-to-fix-utf-encoding-for-whitespaces)). ByteString's `dropWhile` and `dropWhileEnd` only allow to handle single `Char`s.
Data.Text's `strip` ALSO only trims by single `Char`s with `Data.Char.isSpace`, but does it somehow (?) more reliably. I did not have to consider `\194` as a separate case with this solution. That's why I accepted the performance cost of decoding and encoding. Feel free to suggest something else. But what I implemented now seems to fix #213, which should be the minimum goal.


nevrome:
Ok - so @stschiff and I decided to abandon the transformation to `Text` for now. We had a wrong understand of what a `Char` actually is in Haskell. I will prepare another PR with the useful changes introduced here, including the fix for #213.

nevrome:
See PR #220

PR #215, opened 2023-01-24
fixed --no-extract and made it capable to handle non-package entities correctly

nevrome:
I think this should solve #185. It's a very minimal change, actually, but I think this is sufficient.

I was about to write some tests for `--no-extract`, but then I thought I should first test how big the performance increase actually is. A simple test case with 

```
trident forge -d . -f "*2010_RasmussenNature*,*2012_MeyerScience*" -o huhu --no-extract

trident forge -d . -f "*2010_RasmussenNature*,*2012_MeyerScience*" -o huhu
```

showed no relevant difference at all, though:
```
real	0m41,138s
user	0m40,871s
sys	0m0,264s

real	0m41,401s
user	0m41,198s
sys	0m0,201s
```

Do you have a test case where `--no-extract` actually has an advantage, @stschiff? Or was this option only based on the assumption that it *could* be faster? Then I would suggest to deprecate it right away.

The help for `--no-extract` is also a bit misleading:

```
Skip the selection step in forge. This will result in
outputting all individuals in the relevant packages,
and hence a superset of the requested
individuals/groups. It may result in better
performance in cases where one wants to forge entire
packages or almost entire packages. Note that this
will also ignore any ordering in the output
groups/individuals. With this option active,
individuals from the relevant packages will just be
written in the order that they appear in the original
packages.
```

Afaik the individuals in the output files are ALWAYS ordered according to the order in the packages (and the packages in order of their detection by the .yml file detection). Maybe that's something we want to change one day, but it's not entirely clear, what the order should be, given the logic of our query language.


stschiff:
I see OK... well, let's deprecate then. Indeed it was based purely on a feeling. I _might_ have done a test at some point, but can't remember. The parameter space is also relatively large... you can imagine many small packages, or few large ones, and it's not clear to me in which circumstances we want to optimise stuff. So yes, let's deprecate.

stschiff:
OK, so we should add to this PR then the necessary patches that make this option non-available anymore?

nevrome:
Yes - I think this would be OK. Would this be a major release then, because we deprecate a feature? Maybe we just soft-deprecate it with a warning for now.

nevrome:
I had some test run in the background: `--no-extract` IS faster under the right circumstances. Here I merged the current state of the published_data repository.

```
$ trident forge -d . -f "" -o huhu --no-extract
[Info]    SNPs:     10000    36s
[Info]    SNPs:     20000    1m 15s
[Info]    SNPs:     30000    1m 51s
[Info]    SNPs:     40000    2m 26s
[Info]    SNPs:     50000    3m 1s
[Info]    SNPs:     60000    3m 37s
[Info]    SNPs:     70000    4m 15s
[Info]    SNPs:     80000    4m 53s
[Info]    SNPs:     90000    5m 28s
[Info]    SNPs:    100000    6m 4s
[Info]    SNPs:    110000    6m 39s
[Info]    SNPs:    120000    7m 15s
[Info]    SNPs:    130000    7m 52s
[Info]    SNPs:    140000    8m 28s
[Info]    SNPs:    150000    9m 3s
[Info]    SNPs:    160000    9m 39s
[Info]    SNPs:    170000    10m 16s
[Info]    SNPs:    180000    10m 57s
[Info]    SNPs:    190000    11m 36s
[Info]    SNPs:    200000    12m 14s
...
```

```
$ time trident forge -d . -f "" -o huhu
[Info]    SNPs:     10000    53s
[Info]    SNPs:     20000    1m 48s
[Info]    SNPs:     30000    2m 41s
[Info]    SNPs:     40000    3m 35s
[Info]    SNPs:     50000    4m 33s
[Info]    SNPs:     60000    5m 30s
[Info]    SNPs:     70000    6m 31s
[Info]    SNPs:     80000    7m 37s
[Info]    SNPs:     90000    8m 37s
[Info]    SNPs:    100000    9m 34s
[Info]    SNPs:    110000    10m 28s
[Info]    SNPs:    120000    11m 22s
[Info]    SNPs:    130000    12m 19s
[Info]    SNPs:    140000    13m 16s
[Info]    SNPs:    150000    14m 7s
[Info]    SNPs:    160000    14m 56s
[Info]    SNPs:    170000    15m 47s
[Info]    SNPs:    180000    16m 38s
[Info]    SNPs:    190000    17m 30s
[Info]    SNPs:    200000    18m 21s
...
```

So sorry for the uninformed suggestion to deprecate it. I think we should keep it around. But we should document it a bit better and point to a use-case. What do you think?

stschiff:
Yes, OK. I guess the use-case is: If you select a lot of samples from the same package, then `--no-extract` will be faster, since it can simply forward the contents of the entire package in each SNP. I can make a suggestion for the documentation change.

stschiff:
I clarified the help text. I might have overdone it, please have a look @nevrome. Btw, I think `--no-extract` is poor naming. Shall we rename it to `--packagewise` or `--retain_packages` or `--complete_packages`?

nevrome:
Great! The explanation sounds helpful to me! And a new name is a good idea as well. Maybe not in snake_case, because all our other options are in camelCase. I like `--packagewise`.

I'll try to add some tests for the option.

stschiff:
OK, just to be sure: Changing the option to `--packagewise` is a breaking change... but I think hardly anyone has used the `--no-extract` option so far, right?

PR #216, opened 2023-01-24
added a proper warning if one or all baseDirs do not exist

nevrome:
```
trident: gz: getDirectoryContents:openDirStream: does not exist (No such file or directory)
```

is now

```
[Warning] Base directory (-d) "gz" does not exist
```


stschiff (approved):
Nice. Good to go. Thanks.

Issue #128, opened 2021-09-21
Plink format data Phenotype column set to missing for all data


TCLamnidis:
This might be more of an issue the plink formatted packages themselves, instead of poseidon, but I'll put it here for now. 

## Problem description
The plink fam file has the following columns:
> 1: Family ID ('FID')
2: Within-family ID ('IID'; cannot be '0')
3: Within-family ID of father ('0' if father isn't in dataset)
4: Within-family ID of mother ('0' if mother isn't in dataset)
5: Sex code ('1' = male, '2' = female, '0' = unknown)
6: Phenotype value ('1' = control, '2' = case, '-9'/'0'/non-numeric = missing data if case/control)

It seems we generally put/allow column 6 to be all 0s in any(/every) package. Some programs, e.g. `smartpca`, will actually look at this column to decide what individuals to include in the analysis. More specifically, a `0` is interpreted as missing phenotype, and the individual excluded from the run. Since all individuals are set to 0 (as far as I've seen thus far) the following error is thrown, and the runs killed:
```
parameter file: smartpca_runs/poplist1/parameters.txt
### THE INPUT PARAMETERS
##PARAMETER NAME: VALUE
genotypename: forged_package/PCA_package_1.bed    
snpname: forged_package/PCA_package_1.bim        
indivname: forged_package/PCA_package_1.fam        
evecoutname: smartpca_runs/poplist1/PCA_poplist1_plink.evec           
evaloutname: smartpca_runs/poplist1/PCA_poplist1_plink.eval           
poplistname: PCA_poplists/PCA_poplist1.txt
lsqproject: YES     
outliermode: 2      
numoutevec: 4       
## smartpca version: 16000
norm used

lsqproject used
French:HGDP00511 ignored
French:HGDP00512 ignored
French:HGDP00513 ignored
[...]
Adygei:YY-18 ignored
Adygei:YY-36 ignored
all individuals set ignore.  Likely input problem (col 6)
resetting all individual...
genotype file processed
nodata:            rs3094315
nodata:            rs7419119
[...]
nodata:          rs116026001
nodata:        Affx-50065629
number of samples used: 0 number of snps used: 0
Using 1 thread, and partial sum lookup algorithm.
total number of snps killed in pass: 0  used: 0
fatalx:
XTX has zero trace (perhaps no data)
Abort trap: 6
```
After all individuals are set to missing, there is no data left to analyse, and a huge error is printed out, which lists all individuals and SNPs in the dataset!

[...]

PR #222, opened 2023-02-05
Flexible pop name output

stschiff:
- this PR addresses #128 by adding an option to encode popNames flexibly in the plink FAM file.


src/Poseidon/CLI/OptparseApplicativeParsers.hs
```diff
@@ -418,3 +419,16 @@ parseUpgrade = OP.switch (
     OP.long "upgrade" <>  OP.short 'u' <>
     OP.help "overwrite outdated local package versions"
     )
+
+-- PlinkPopNameAsFamily always is the default
+parsePlinkPopMode :: OP.Parser PlinkPopNameMode
+parsePlinkPopMode = parsePlinkPopAsPhenotype <|> parsePlinkPopAsBoth <|> pure PlinkPopNameAsFamily
```
nevrome, line 425: I think I would prefer one option with three possible iterations of an argument (e.g. `--plinkPopName asPhenotype | asBoth | asFamily`). It's more flexible in case other options emerge in the future (unlikely, I know, but still), fits to the other options we already have (e.g. `--outFormat`) and, most importantly, is less verbose and repetitive in the CLI documentation.
image
But I understand that this is a matter of taste.

src/Poseidon/CLI/Forge.hs
```diff
@@ -189,13 +194,13 @@ runForge (
     currentTime <- liftIO getCurrentTime
     newNrSNPs <- liftIO $ catch (
         runSafeT $ do
-            (eigenstratIndEntries, eigenstratProd) <- getJointGenotypeData logEnv intersect_ relevantPackages maybeSnpFile
-            let newEigenstratIndEntries = map (eigenstratIndEntries !!) relevantIndices
-
+            (eigenstratIndEntries, eigenstratProd) <- getJointGenotypeData logEnv intersect_ plinkPopMode relevantPackages maybeSnpFile
+            let eigenstratIndEntriesV = eigenstratIndEntries
```
nevrome, line 198: What's the meaning of this suffix `V` variable? I remember that I removed that recently. Now it's back. What am I missing here? Is this some clever design pattern? :sweat_smile:

src/Poseidon/CLI/Init.hs
```diff
@@ -47,7 +50,7 @@ runInit (InitOptions (GenotypeDataSpec format_ genoFile_ _ snpFile_ _ indFile_ _
     -- create new package
     logInfo "Creating new package entity"
     outName <- liftIO $ determinePackageOutName maybeOutName outPath
-    inds <- liftIO $ loadIndividuals outPath genotypeData
+    inds <- liftIO $ loadIndividuals outPath genotypeData plinkPopMode
```
nevrome, line 53: General observation: Maybe we will need a general data type for these genotype data settings like `plinkPopMode`. I can for example imagine that more settings may emerge if we ever come around to allow .vcf files in Poseidon packages.

nevrome (commented):
Looks good :+1:
Maybe we can add some small tests to see the behaviour in action. But beyond that and my concerns about the interface I think this can be merged.

src/Poseidon/CLI/OptparseApplicativeParsers.hs
stschiff: No I agree, actually. Good point. I'll change this.

src/Poseidon/CLI/Forge.hs
stschiff: Oh, I think that may have creeped in during the merge. I also can't remember. I know I usually use the `V` suffix to denote `Vector`s and distinguish them from lists. Not sure what that has to do with this place. Hmm, happy to simplify this.

src/Poseidon/CLI/Init.hs
stschiff: Yes, well I was hoping we could creep that into `GenotypeDataSpec`, but unfortunately that type maps 1 to 1 to the YAML, so that won't work.

nevrome: Ah - I see. I think for now it's fine to keep it as a special option. We can generalize it later when the need arises.

stschiff: Agreed.

src/Poseidon/CLI/OptparseApplicativeParsers.hs
stschiff: Done.

src/Poseidon/CLI/Forge.hs
stschiff: Done.

stschiff:
Please one last look, also regarding the version bump and changeling, @nevrome, then I'll merge. I will not prepare a release for now, as we enough in the pipeline right now.

nevrome:
I quickly wanted to test this and ran into something that's conceptually unclear to me. `--plinkPopName` is a single option that specifies both the in- and the output. What happens if somebody wants to *change* the way group names are encoded in a plink genotype dataset? Wouldn't that be the most natural application of this feature, for example to make a dataset fit into a tool that requires either group name encoding? 

Why did you decide to only add  `--plinkPopName` and not  `--inPlinkPopName` and `--outPlinkPopName`?

Thinking even further about this, I now finally understand that this should indeed go into `GenotypeDataSpec`, if we wanted to be correct. This is a per-dataset setting. And that would mean we need a change to the schema. We could squeeze it into 2.7.0. Just another optional field in the .yml file.

But I guess you already saw all of this and decided against a schema change. I think I could also be convinced not to go there for this rather obscure feature. But a user should at least have the option to "fix" a dataset to read it together with others. And for that we need the distinction of in- and output, right?

stschiff:
Hmm, crap. You're right. 

OK, that requires more thinking on the interface then. I would definitely like to avoid storing that obscure thing in the YAML file. It's also not actually necessary: Given that we know the group name already from the Janno file, we can really infer the encoding easily from comparing them group names in the Janno file with the Family- and Phenotype fields of the fam file.

I suppose one could then interpret this option only for writing? and rename it to `--outPlinkPopName` and simply infer the input one?

Alternatively a global `--inPlinkPopName` and `--outPlinkPopName` option might also do the job. What do you think?

nevrome:
I fear we can not rely on the .janno file because we allow reading genotype data without a .janno file.

I wonder if we could instead infer the setting from the .fam file itself. What is in the respective columns, when they are not (mis)used for group names? Single digit integers, right? Maybe a logic like this would do:

1. If (first col == last col) then assume `asBoth`.
2. If (first col is multi letter character & last col is single digit integer) then assume `asFamily`
3. If (first col is single digit integer & last col is multi letter character) then assume `asPhenotype`
4. Else assume `asFamily`

Or something like this. Could be expressed even simpler. And then we can pair this with `--outPlinkPopName` with a default of `asFamily`.

stschiff:
To summarise a quick phone discussion: We'll opt to have one global option `--inPlinkPopName` and one global option `--outPlinkPopName`. That's the cleanest solution for now that requires the least code changes and maximises transparency to the user. I'm on it.

stschiff:
OK, so big overhaul. First, as promised, we now distinguish between `--inPlinkPopName` and `--outPlinkPopName`. Second, thanks to forcing `defaultPackageReadOptions` to explicitly expect the Input-PlinkMode I identified tons of commands, and the server, where I had forgotten to take in that option. Now that I have changed that, I can see that indeed _all_ commands require this option. This makes it a basic option, and in principle it therefore should go in the CLI interface to a higher level than the individual CLI command options, similar to `--logMode` and `--errLength`. 

I have not done this yet, as it would also best be combined with a change in our basic Monad stack, which already is a `ReaderT`. Currently, we only use it to pass along the Logger, but if we have such a general and basic setting like `--inPlinkPopName`, we could reduce verbosity by changing the basic reader environment to not only feature the Logger, but also other settings.

We currently have:

```
type LogEnv = LogAction IO Message
```

And we could change this to:

```
data Env = Env {
    _envLogAction = LogAction IO Message
    _envInputPlinkMode = PlinkPopNameMode
}
```

This would then also mean that we change `usePoseidonLogger :: LogMode -> PoseidonLogIO a -> IO a` to `usePoseidonLogger :: Env -> PoseidonLogIO a -> IO a` which is trivial.

In `logMsg`, we would change `LogAction logF <- asks getLogAction` simply to `LogAction logF <- asks (getLogAction . _envLogAction)`

In addition, we could simply do `plinkMode <- asks _envInputPlinkMode` to retrieve the plink mode from anywhere within the monad.

OK, so we don't have to do that right away, what I have committed in this PR now should work, but it is a bit annoying that every single CLI command now expects this one option, which requires to be passed on to all main functions.

stschiff:
Let me prepare a little experimental PR on top of this PR, so we can see how this would look. I'm open to not doing it! But it would open us to taking in more basic options in the future, which would be rather convenient.

Another question is to what extent our current poseidon-hs library already depends on the `PoseidonLogIO` Monad Stack. Sadly it is already used inside `Janno.hs`, `Package.hs` and `GenotypeData.hs`.

Ideally, if we want to make the library more widely useable, we might want to use a TypeClass to make this more independent. But - alas - if our library functions want Logging and access to general Poseidon settings, perhaps using this basic Monad Stack is just something that clients of our library will have to live with.

stschiff:
OK, I love it. And I ended up committing to this branch. I think you will see that by extending the base monad (now called `PoseidonIO`, we have almost no change in many of the CLI modules. We basically now have a Monad that provides an environment `Env` with:
* Logging, via a LogAction (now called `LogA` instead of `LogEnv`)
* Basic settings, currently only `PlinkPopNameMode`

All functions within that monad have access to these two things, so this frees us from passing them down to individual functions and declutters the code quite a bit compared to my previous commit.

Have a look, @nevrome, I think this is the way to go about general settings such as this one.

nevrome:
I love it! Just two things:

1. It would be good to have some tests for the new options. Maybe a golden test cycle where a non-default package is created and read again.
2. The command line help for `trident` became very verbose with this long text for `--inPlinkPopName`. I think this makes it harder to have a "conversation" with `trident`. I know that optparse-applicative is somewhat opinionated about the way it generates the help text, but maybe we can hack it a bit to hide this wall of text. Or we just shorten the text and rely on the website for verbose documentation.

stschiff:
Nr 2 is done. Now implementing some tests

stschiff:
OK. Tests added. Please have a quick look, @nevrome, as I'm never sure I'm handling the golden test framework correctly. If you think it's OK, feel free to merge this. I will also be back online later.

nevrome (approved):
Looks good to me :+1:

stschiff:
OK, then I'll merge this.

Issue #225, opened 2023-02-08
`-j` in `list` should support arbitrary, additional columns


nevrome:
The foundation for that is already implemented in #221 with relevant `ToJSON` and `FromJSON` instances. Ideally this should include a refactoring of the verbose and crude `extractAdditionalField`.

PR #221, opened 2023-02-05
Janno input validation for JSON and CSV data

nevrome:
In this PR I collect the changes for #167.

Here's some test code to check if en- and decoding to JSON works as expected:

> (decode $ encode $ JannoRow {jPoseidonID = "1", jAlternativeIDs = Nothing, jRelationTo = Nothing, jRelationDegree = Nothing, jRelationType = Nothing, jRelationNote = Nothing, jCollectionID = Nothing, jSourceTissue = Nothing, jCountry = Nothing, jLocation = Nothing, jSite = Nothing, jLatitude = Nothing, jLongitude = Nothing, jDateC14Labnr = Nothing, jDateC14UncalBP = Nothing, jDateC14UncalBPErr = Nothing, jDateBCADMedian = Just $ BCADAge 191, jDateBCADStart = Nothing, jDateBCADStop = Nothing, jDateType = Just C14, jDateNote = Nothing, jNrLibraries = Nothing, jCaptureType = Just $ JannoList [Shotgun, TwistAncientDNA, A1240K], jGenotypePloidy = Just Diploid, jGroupName = JannoList {getJannoList = ["huhu"]}, jGeneticSex = JannoSex Female, jNrSNPs = Nothing, jCoverageOnTargets = Nothing, jMTHaplogroup = Nothing, jYHaplogroup = Nothing, jEndogenous = Nothing, jUDG = Nothing, jLibraryBuilt = Nothing, jDamage = Nothing, jContamination = Nothing, jContaminationErr = Nothing, jContaminationMeas = Nothing, jContaminationNote = Nothing, jGeneticSourceAccessionIDs = Nothing, jDataPreparationPipelineURL = Nothing, jPrimaryContact = Nothing, jPublication = Nothing, jComments = Nothing, jKeywords = Nothing, jAdditionalColumns = CsvNamedRecord (HM.fromList [("de", "ab"), ("gugu", "45")])}):: Maybe JannoRow

PR #226, opened 2023-02-14
Refactoring `list --individuals -j ...`

nevrome:
I'm either a fool or I had a stroke of genius. Difficult to judge with this one.

As suggested in #225 I replaced the unsatisfying `extractAdditionalField` with a much simpler, much more flexible solution, that relies on the hashmap in Data.CSV. `JannoRow` already is an instance of `Csv.ToNamedRecord`, because that's the mechanism we use to write .janno files. I think we can abuse this fact to transform `JannoRow`s to hashmaps and read their content as strings for arbitrary keys (so: .janno column names). What do you think?

A big problem is, that we can not really test this with the remote server at the moment. My changes in 1.1.8.4 (#221) changed the way janno data is encoded in JSON, so server and client can not speak to each other any more. `--remote` triggers an error, when I try to read from the old server with the new client:

```
[Error]   Error in parsing JSON: "Error in $[0][1][0].jGenotypePloidy: Genotype_Ploidy \"Haploid\" not in [diploid, haploid]"
```

This is a separate issue from this PR, but it affects how I can test it. See #228 for an attempt to fix that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- V 1.1.8.6: Refactored the `-j` mechanism by which `list --individuals` includes additional variables in the output table. It is now possible to query arbitrary addititional columns
 - V 1.1.8.5: Rolled back some of the ToJSON instances changed in 1.1.8.4 because they broke backwards compatibility of the server-client communication. Added some additional tests to prevent such oversights in the future. Slightly reorganized the golden tests
 - V 1.1.8.4: Unified the implementation of ToJSON/FromJSON and ToField/FromField instances for .janno datatypes to perform input validation through smart constructors
 - V 1.1.8.3: The fix in introduced in 1.1.8.1 introduced a bug: It broke valid unicode characters in .janno files and prevented reading them. The solution implemented here solves this issue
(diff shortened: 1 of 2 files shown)


stschiff (commented):
OK, yes, I think this is an obvious improvement! Thank you. I have two comments:

1) Is it correct that now - since `extractAdditionalFields` has become pure, we cannot anymore output a warning if a field is requested that is not there, not even in the extended Janno with additional non-schema columns? I think that should definitely be considered warning-worthy!

2) Right now, if it doesn't find a field, it doesn't output `n/a` but an empty string. Is that what we want?

nevrome:
1. Yes, that is true. I thought it's not important to print a warning here, actually. If a user requests a non-existent column, then they get an empty column. But alright - I think it should be possible to add one.
2. I thought that is fine, because we treat empty strings and "n/a" equally in the .janno file. But I guess I can change that as well.

stschiff:
OK, for 1) yes, I think a warning would be good, at least if a column is requested that is present nowhere. For 2) OK, that's fine then.

PR #228, opened 2023-02-15
fix server-client interaction broken since 1.1.8.4

nevrome:
v1.1.8.4 (#221) broke the communication between old (before 1.1.8.4) versions of `poseidon-http-server` and new versions of `trident`. This was foreseeable, yet unforeseen

This PR is meant to fix this incompatibility by reverting some of the JSON encoding/decoding related changes. This is not entirely trivial, because we do need a working JSON scheme for `CsvNamedRecord` (which did not exist in the past). To make sure that everything works, we need to consider the following cases:

- [x] 1. Old server + New client
- [x] 2. New server + Old client
- [x] 3. New server + New client

I ran some quick tests on the command line with a local server, and old and new client versions.
```bash
# v1.1.8.5 client + old (?) server
trident list --remote --packages | --groups | --individuals  -j Nr_SNPs

#  v1.1.8.5 client +  v1.1.8.5 server
poseidon-http-server -d 2010_RasmussenNature -c -z testDir
trident list --remote --remoteURL "http://127.0.0.1:3000" --packages | --groups | --individuals -j Nr_SNPs

#  v1.1.7.0 client + v1.1.8.5 server
wget https://github.com/poseidon-framework/poseidon-hs/releases/latest/download/trident-Linux
chmod +x trident-Linux
./trident-Linux list --remote --remoteURL "http://127.0.0.1:3000" --packages | --groups | --individuals -j Nr_SNPs
```
Works for all combinations.


nevrome:
OK - I'm merging this now. Further testing of this very change will happen in #226. I'm taking the full responsibility!

Issue #229, opened 2023-02-17
Make `[JannoRow]` an instance of `Monoid`


nevrome:
A janno file is currently defined as `[JannoRow]`. To combine two janno files we have a dedicated function
```haskell
combineTwoJannos :: [JannoRow] -> [JannoRow] -> [JannoRow]
```
This is necessary to keep proper track of arbitrary, additional columns. We also have a dedicated function to concatenate janno files, which makes use of `combineTwoJannos`. 
```haskell
concatJannos :: [[JannoRow]] -> [JannoRow]
concatJannos = foldl' combineTwoJannos []
```
Is is easy to forget to use these functions, though, because `Data.List.concate` will of course also typecheck. I almost ran into this issue yesterday.

Can we prevent this by making `[JannoRow]` an instance of `Monoid`? We should probably give it a proper name in the process, maybe `JannoFile`.

I think we only have to implement `<>`/`mappend` for `Semigroup` (which is essentially `combineTwoJannos`) and `mempty` which should be `[]`. We could also overwrite the default `mconcat` (`foldr mappend mempty`) with the potentially more efficient `foldl' mappend mempty`, which is what we currently have in `concatJannos`.

I feel like we already discussed sth. like this in the past, @stschiff. I hope I didn't forget a crucial drawback of this idea.

PR #232, opened 2023-02-18
Added a newtype `JannoFile` for `[JannoRow]`

nevrome:
Here's an implementation of the idea shared in #229


stschiff (approved):
OK, I approve this. I think personally I would have not replaced all instances of `[JannoRow]` with `JannoFile`, in particular if the function just uses the rows directly, in which case I would have simply left things as they are. The monoid instance is useful, but only for merging Jannos, so it doesn't require making this change all the way through.

But I certainly don't mind how it is now, so this is good to merge.

nevrome:
Right - my biggest concern is that `JannoFile` doesn't really capture the meaning. It's just an arbitrary list of `JannoRow`s. I considered naming it `JannoSet`, but a set has no order. `JannoList` is already in use. Do you have a suggestion?

I replaced all instances of `[JannoRow]` to make sure that it becomes harder to accidentally run into the issue I describe in #229. But I understand that it looks a bit weird at times, not least because of the type name.

stschiff:
Yes, the name isn't perfect. You could have used `JannoMonoid`, but that's not a name you want everywhere. What about `JannoRows`?

nevrome:
`JannoRows` is simple and clear. I love it. Will use that. Thanks!

PR #239, opened 2023-03-22
Generic survey

nevrome:
This is an idea how to remove the boilerplate from the Survey module. It's advanced magic I barely understand. But it's also super cool and it doesn't tank the compile time. It also shouldn't change the perceived behaviour of `trident survey`. What do you think, @stschiff?

To make this work as expected I had to adjust the order of the fields in the JannoRow record type.


stschiff (approved):
OK, I've spent an appreciable amount of time on this, and I think I give up to some extent to understand this properly. But I think since it's a change quite limited in scope, and I trust that it satisfies our golden Tests and perhaps your manual testing, I'm green-lighting this. 

Impressive, this type-level mapping stuff. Perhaps you can explain to me at some point those bits that you do understand (it must be more than I do, as I would not have been able to write this from the Stackoverflow answer alone).

OK to merge! And Thanks!

PR #242, opened 2023-04-04
started upgrading to lts-20.11

stschiff:
I'm just putting this here for now. The build plan works, but we needed to get rid of yaml-pretty-extras. The newer `yaml` packages support tweaking the output order, though: https://stackoverflow.com/a/65659340

So that's something to be changed.


stschiff:
Wow, I did only just see this. Fantastic. Then I can compile this on my new Mac as well. I will try and green-light this tonight

stschiff (commented):
Thanks. I'll merge this after a compilation run on my Macbook

PR #245, opened 2023-04-28
Changes for Poseidon v2.7.1

nevrome:
Working though the validation of real world .ssf files provided by @93Boy in https://github.com/poseidon-framework/published_data/pull/114/ exposed various flaws in trident and the Poseidon schema (e.g. https://github.com/poseidon-framework/poseidon2-schema/issues/61). In this PR I will collect the changes necessary to trident in anticipation of changes to the schema.


src/Poseidon/CLI/Forge.hs
```diff
@@ -255,7 +256,7 @@ filterSeqSourceRows (JannoRows jRows) (SeqSourceRows sRows) =
     where
         hasAPoseidonID :: [String] -> SeqSourceRow -> Bool
         hasAPoseidonID jIDs seqSourceRows =
```
stschiff, line 258: I guess that should be singular? `seqSourceRow` as the more correct variable name?

src/Poseidon/SequencingSource.hs
stschiff, line 1: Very cool what's going on here with the accession IDs!

src/Poseidon/SequencingSource.hs
```diff
@@ -273,13 +396,17 @@ readSeqSourceFile seqSourcePath = do
     -- error case management
```
stschiff, line 396: Just a few lines up, in L 391: Cleaning out quotation marks from header names is fine, but technically might break things if people (bizarrely) would put escaped quotes into header strings right? I think we can accept that.

src/Poseidon/SequencingSource.hs
```diff
@@ -293,23 +420,30 @@ readSeqSourceFileRow seqSourcePath (lineNumber, row) = do
                 Right (pS :: SeqSourceRow) -> do return $ Right pS
 
 -- SeqSource consistency checks
-
 checkSeqSourceConsistency :: FilePath -> SeqSourceRows -> Either PoseidonException SeqSourceRows
-checkSeqSourceConsistency seqSourcePath xs
-    | not $ checkSamplesUnique xs = Left $ PoseidonFileConsistencyException seqSourcePath
-        "The values in the sample_accession column are not unique"
+checkSeqSourceConsistency _ xs
+    -- | ... no tests implemented
```
stschiff, line 425: OK, I suppose this is now empty because of the lack of unique keys, and lack of explicit foreign-key constraints? Do you want to keep those functions for later purpose?

test/Poseidon/SequencingSourceSpec.hs
```diff
@@ -44,16 +48,31 @@ testReadSeqSourceFile = describe "Poseidon.SequencingSource.readSeqSourceFile" $
...
         (SeqSourceRows s) <- testLog $ readSeqSourceFile normalFullSeqSourcePath
         length s `shouldBe` 3
-        map sPoseidonID s                `shouldBe` [JannoList ["Ash033.SG"], JannoList ["Ash002.SG"], JannoList ["Ash040.SG"]]
+        map sPoseidonID s                `shouldBe` [Just $ JannoList ["Ash033.SG"], Just $ JannoList ["Ash002.SG"], Just $ JannoList ["Ash040.SG"]]
         map sUDG s                       `shouldBe` [Just SSFMinus, Just SSFHalf, Just SSFPlus]
         map sLibraryBuilt s              `shouldBe` [Just SSFSS, Just SSFDS, Just SSFDS]
-        map sGeneticSourceAccessionIDs s `shouldBe` [INSDCBioSample "SAMEA7050454", INSDCBioSample "SAMEA7050404", INSDCBioSample "SAMEA7050455"]
-        map sStudyAccession s            `shouldBe` [Just $ INSDCProject "PRJEB39316", Just $ INSDCProject "PRJEB39316", Just $ INSDCProject "PRJEB39316"]
-        map sRunAccession s              `shouldBe` [Just $ INSDCRun "ERR4331996", Just $ INSDCRun "ERR4332592", Just $ INSDCRun "ERR4332593"]
+        map sSampleAccession s           `shouldBe` [ Just $ AccessionIDSample $ INSDCBioSample "SAMEA7050454"
```
stschiff, line 54: Super cool!

stschiff (approved):
Look all very good. Some minor comments/questions left in. What do we do about the flagged lines that aren't covered by tests yet? Leave them as they are right now to move forward given the limited developer time we have? Or strive towards higher coverage? I'm fine with leaving it like this for now.

src/Poseidon/SequencingSource.hs
nevrome: Yes - I thought it would be fine to keep them around for the future.

src/Poseidon/CLI/Forge.hs
nevrome: True. Fixed :+1:

src/Poseidon/SequencingSource.hs
nevrome: I copied this from the .janno code. Might indeed have weird effects in edge cases. I added a comment now to warn of the danger.

nevrome:
I didn't test any of the potential error states. For the janno code we have some tests for these, but also no fully comprehensive system. I think it's OK to leave it like that for now, but my arguments are weak:
- I ran some tests on the command line
- Writing hundreds of tests is boring
- I trust the type- and parsing system

Maybe we do a test-hackathon one day to get to 90+%

PR #255, opened 2023-06-07
reworked API field names

nevrome:
A suggestion for more consistent API field names.
image
image
image

### DIFF
--- a/src/Poseidon/SecondaryTypes.hs
+++ b/src/Poseidon/SecondaryTypes.hs
@@ -83,16 +83,16 @@ instance HasNameAndVersion ExtendedIndividualInfo where
 instance ToJSON ExtendedIndividualInfo where
     toJSON e =
         object [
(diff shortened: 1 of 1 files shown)


stschiff (approved):
OK, I like the new naming.

I can see why you wanted to flatten the GroupInfo JSON instances. It's OK for me, but then you should remove the JSON instances from `PacNameAndVersion`, as that type is not anymore used for JSON serialization.

Otherwise I approve this.

nevrome:
> OK, I like the new naming.
> 
> I can see why you wanted to flatten the GroupInfo JSON instances. It's OK for me, but then you should remove the JSON instances from `PacNameAndVersion`, as that type is not anymore used for JSON serialization.
> 
> Otherwise I approve this.

Done.

stschiff (approved):
Looks good, thanks!

Issue #250, opened 2023-05-19
Add a global option `--ignoreChecksums` to trident


nevrome:
Would have saved a lot of time during testing and debugging in the past. Of course it's technically only relevant for sub-commands that read packages, but that also applies to `--inPlinkPopName`. So I think a global option is the way to go here.

stschiff:
Yep, good idea and easy to do. Should we perhaps not document it but treat as an internal option only? Why would average users want to do this? Perhaps if they quickly want to do something and forgot to update the checksums?

nevrome:
Well - I see no reason to omit this option. The checksums can be a pain for everybody and sometimes it's convenient to just ignore them, e.g. when modifying a package.

Which does not mean I want to get rid of them: I feel they are useful for reviewing/monitoring changes in our public repository. And they fulfill a very important second purpose: Download validation.

nevrome:
Closed in #263.

PR #263, opened 2023-08-02
Reworked update into rectify

### LINKED ISSUE #164: Rename the update subcommand
The update subcommand was designed to support the work with the public repo. Its purpose is to rectify some technical aspects of poseidon packages after manually (!) updating them. It does not adjust the structure of packages from old to new versions of the Poseidon schema.

As predicted by you, @stschiff, this is not what users expect. @arthurKocher contacted me exactly about this confusion. Maybe we should discuss two options:

- Should we rename `update` to reduce this confusion? What would be a clear yet concise verb to describe what `update` currently does?
- Should we think about implementing a "real" `update` command, that takes packages in old schema versions and automatically brings them to the latest (or even any newer/other) version? It would be some work, yet by no means impossible. But is this a good investment of our time?

### LINKED ISSUE #249: Make `update` more modular
I think it would be very useful to be able to control the behavior of `trident update` more precisely. The idea is a bit similar to #248.

At the moment `update` is a catch-all procedure where the user can exclude certain steps (with `--noChecksumUpdate` or `--ignoreGeno`). I think we should turn this around and make it **opt-in**. Each aspect of a package that should be updated should also be named explicitly in the call to trident update.

That may translate to an interface like this:

```
trident update (-d|--baseDir DIR)
    [--poseidonVersion ARG] # this is already opt-in
    [(--packageVersion ARG) (--changelog ARG)]
    [--checksumGeno] [--checksumJanno] [--checksumSSF] [--checksumBIB] [--checksumAll]
    [--newContributor ARG]
    ...
```

This rewrite should be paired with a rename as proposed in #164.

nevrome:
When merged this will close #249 and close #164.


nevrome:
Merging this will also close #250.

PR #265, opened 2023-08-11
Integrate xerxes

stschiff:
Still a draft, but integration was fairly easy. I will work on some tests now.


stschiff:
See also https://github.com/poseidon-framework/poseidon-analysis-hs/pull/24

stschiff:
I removed the .Rhistory file. It was in fact historic.

I think Xerxes will follow the same versioning now as trident. That is - in my view - the condition that comes with merging it into one repository. Otherwise we have to introduce a separate versioning from the one in the cabal-file, and I fear that will get quite obfuscating.

Regarding the major version number, I agree. I have no opinion on the era-number, though I am a little afraid it might cause all too large expectations 😅

stschiff:
OK, I have increased the first version number now, let's be bold.

Regarding the tests, just wanted to say that for FStats we have to do some manual tests, since I want to make sure that the results are similar to what previous tools, including Nick's admix tools gives. Those runs are too long to automate.

Let me know if there is anything else.

stschiff:
We'll close this for now. Was not a good idea for now.

Issue #230, opened 2023-02-17
Implement a way to specify the output order of individuals in `forge`


nevrome:
@AyGhal and I realized that it would be very helpful if one could set the output order of individuals in `forge`.

Currently the order depends on the order in which packages/data are discovered by `-d` and `-p`. This is not ideal. 

Maybe this could be implemented by considering the entity order in the `--forgeString` DSL.

stschiff:
In full generality, this would require a complete rewrite of how we currently read the forge DSL. With the recent additions implemented in #211 I must admit I'm tempted to actually do this. We started off with this cool idea of identifying forge-samples with a single-pass through the individuals-list. This is clever and fast, but it also means that the output order will always be determined by the input order.

If we want to the output-order to follow the forge-file, then we need to switch from our current algorithm to a kitchen-chef-algorithm. Meaning that instead of going through the individuals-list and choosing what's in the forge-file, we need to go through the forge-file instead and choose from the individuals whatever is needed. 

I think this will technically be slower (because usually the individuals-list will be much longer than the forge-list), but perhaps only virtually so, since this is all super-fast on today's CPUs. 

So technically, it's a simple algorithm:
```
for each forge_item in forge_list; do
   go through individuals-list and select out the items that match the forge_item
done
```

This would then also make the code added in #211 quite a bit simpler!

nevrome:
Phew... I would prefer not to rework this crucial code yet again, especially as it's relatively well tested now and more pressing tasks are at hand. I'm also not convinced yet that a full rewrite is necessary.

Maybe some changes to the entity extraction algorithm in and around `conformingEntityIndices` would already be sufficient.

stschiff:
I would welcome that change at some point in the future, and I would volunteer to do a rework. I think already with the new priority syntax, the old single-pass code has become difficult to read and maintain and I would anyway prefer a kitchen-chef algorithm. But I admit that it's something for the future.

nevrome:
This feature was added in #301

PR #14, opened 2020-11-27
Core library refactoring

stschiff:
Major refactoring of code in Poseidon.Packages into separate submodules. Also cleaned up exception handling, as a suggestion.

PR #268, opened 2023-08-29
Updates stephan

stschiff:
I am working on a few updates and some refactoring of #264


stschiff:
Alright, this has become very substantial, and I will document as much as I can what has happened.

# High-level overview
I realised that much of the complexity in the `EntitiesList` module was caused by the fact that our entity-conformity algorithm suffered from what I would call non-locality of entity specifications, which in my view is undesirable. Specifically, I identified two sources of such non-locality that made entities inherently non-selfconsistent.

The first source of interaction came from the complex forge-semantics, which I lay out in detail in #269 and which I explicitly change in this PR (I suggest to document the proposed semantics if we agree).

The second source of interaction came from the fact that in many entity-selection semantic clauses, we need to have knowledge on whether a given individual belongs to the _latest_ package. Due to our data structure, specifically `IndividualInfo`, we did not have access to that knowledge at the time when we make the entity selection, thus necessitating a two-step selection process which became pretty intricate.

I address both of these issues in this PR, which by the way lives as branching off of #264, simply to get more specific diffs, but we could also opt to merge this into master after review and revision, of course.

# Refactoring IndividualInfo and PoseidonEntity
I realised that much of the necessary refactoring was difficult because one of our key data structures, `IndividualInfo` became entangled in two quite different roles, which had conflicting goals. One role was in `PoseidonIndividual`, namely to specify a certain _query_ of a specific individual. The other role was to specify an _actual_ individual in our data base. The first thing I did, therefore, is to disentangle these two roles. Specifically, I redesigned `PoseidonEntity` to now be flat, like this:

```
data PoseidonEntity =
      Pac PacNameAndVersion -- ^ all individuals in a package. A version can be specified, if not implicitly request the latest
    | Group String          -- ^ all individuals with a given group, in all of the latest packages
    | Ind String            -- ^ all individuals with the given name, in all of the latest packages
    | SpecificInd String String PacNameAndVersion -- ^ the individual specified by its name, group and package. If not versioned, then take the latest version.
    deriving (Eq, Ord)
```

Once decoupled, `IndividualInfo` was free to evolve further. Specifically, I now added an `indInfoIsLatest :: Bool` field, which would now store whether the given individual would belong to the latest package. 

## Side point: Server-Client types
As a side-point, I then also remembered that I created `ExtendedIndividualInfo` specifically because I could not repurpose `IndividualInfo` at the time because of its double role. Now, I removed `ExtendedIndividualInfo` and instead added the property of additional janno columns right into `IndividualInfo`, which reduced some redundancy. Note that I also moved `PackageInfo` and `GroupInfo` into the `ServerClient` module, as they are solely used for that (unlike `IndividualInfo`, which lives in `EntityTypes`.

Note that I kept the same JSON instances for `IndividualInfo` as before, to keep the server-client communication backwards-compatible. This then also meant that `infInfoIsLatest` is not something that the server can transmit (unfortunately). Instead, the server just fills this all with `False`, and we use the `setPacVersionLatest` from `EntityTypes` to compute this property for a given list of `IndividualInfo`s accordingly after we retrieve such a list from the server or from local package archives (see also `getJointIndividualInfo` from `Package`). 

# Drastically simplifying EntitiesList

With these changes, much of the core code in EntitiesList became unnecessary. I removed the non-trivial `Eq` instances (which I found very scary actually, assuming that we could easily be surprised by `(==)` not doing exactly what it should do, i.e. comparing structural equality). 

The one single point of complexity now is all in `indInfoConformsToEntitySpec` for `SignedPoseidonEntity`, which literally lists all the 12 cases that I layed out in #269.

The whole entity-resolve stuff basically got reduced to a few lines of code, now that we don't need any post-processing of the selection. Here is the new type of `resolveEntityIndices`:

```resolveEntityIndices :: (EntitySpec a) => [a] -> [IndividualInfo] -> [Int]```

Because the resolution step is now a one-stop show, I wrote a separate function called `reportDuplicateIndividuals`, which simply finds duplciate individuals and makes suggestions for queries using `SpecificInd`, that is now being used in `forge` to help the user in case of duplicates. (Note that duplicates are now always unresolved, according to the new semantics. There is no concept of "resolved duplicates". It's all in the user's responsibility to use negative selection to resolve them.)

```reportDuplicateIndividuals :: [IndividualInfo] -> [(IndividualInfo, PoseidonEntity)]```

this then gets called in `forge` to report to the user that duplicates were found and how they could be resolved.

# Mini-point: List-output including "latest"

Having access to the `isLatest` property in `IndividualInfo`, I now print this out in `list` as "latest" in all cases where this is the case (which includes singleton packages with only one version... to be debated).

**Comment by nevrome** [comment]
That sounds like an all around very good solution! Your detailed analysis of the forgeScript semantics was very inspiring and a great learning experience for me. `indInfoConformsToEntitySpec` is very beautiful now.

That being said, here are some comments in no particular order:

- Removing the double-purpose of `IndividualInfo` sounds very sensible. I'm a bit confused why you decided right away to give it yet another double role as a replacement for `ExtendedIndividualInfo`, though? `indInfoAddCols` and the `infInfoIsLatest = False` hack are a pretty clear indication that this change does not come naturally, and that `IndividualInfo` is again shoehorned into two roles. I think it would be better to be consistent and keep `ExtendedIndividualInfo` as a separate type for the server-client communication. Maybe even with a more explicit name.
- For one particular case, the new selection language semantics are a significant downgrade. Imagine a set of `N` packages with `M` samples that all overlap in their Poseidon_ID. Now imagine we want to select an arbitrary permutation of these `M` samples from the `N` packages. With the previous semantics this would have been possible with `M` positive entries to the forgeScript file. Now it requires `(N-1)*M` negative entries, if I'm not mistaken. The explicit de-selection you require makes this edge case very unpleasant. I hope this is indeed an edge-case.
- I originally hoped that we could make the version selection feature a non-breaking change, but I understand your reasoning as laid out here and in #269. We will have to increase the major version number yet again.
- This breaking change requires a good deal of documentation, both for the trident changelog and for the website. Maybe we can split the workload a bit... We can certainly reuse some of the text you wrote for the issue.
- When we're already breaking things: How difficult would it be to order the output according to the entities in the forgeScript now? With your massive simplifications (I'm jealous of your genius,  when I look into this part of the code, btw.) it seems like we would only have to make some adjustments to `resolveEntityIndices` and `indInfoConformsToEntitySpecs` to get this functionality. Just curious what you think.

With these general comments out of the way I will now look into the code. I especially want to see if my understanding of the forgeScript semantics are reflected in the tests' behaviour.

**Review by nevrome — COMMENTED** [review]
About half-way through - have to take a break now and will continue later.

All good, but I saw multiple examples where merging `IndividualInfo` and `ExtendedIndividualInfo` caused weird type signatures. Some functions now have to be called with a trailing `[]`, because they fulfill different purposes under different circumstances. I don't like this design.

I'm also not a big fan of your excessive use of the list monad. It's elegant, but hard to read. I guess that comes down to personal preferences.

File: test/Poseidon/EntitiesListSpec.hs [hunk 1]
```diff
@@ -186,34 +190,35 @@ testPacReadOpts = defaultPackageReadOptions {
 testBaseDir :: [FilePath]
 testBaseDir = ["test/testDat/testPackages/ancient"]
 
-goodEntities :: EntitiesList
+goodEntities :: SignedEntitiesList
```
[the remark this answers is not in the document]
nevrome: Ah - to test inclusion and exclusion in the same go. Ok.

test/Poseidon/EntitiesListSpec.hs
```diff
@@ -233,160 +238,130 @@ testResolveEntityIndices =
...
-                    Just (Pac (PacNameAndVersion "Pac1" Nothing))
-            decode "\"-<Ind1>\""  `shouldBe`
-                    Just (Exclude (Ind (SimpleInd "Ind1")))
-            decode "\"-<c:b:a>\"" `shouldBe`
-                    Just (Exclude (Ind (SpecificInd $ IndividualInfo "a" ["b"]  (PacNameAndVersion "c" Nothing))))
-            decode "\"-Group1\""  `shouldBe`
-                    Just (Exclude (Group "Group1"))
-            decode "\"-*Pac1*\""  `shouldBe`
-                    Just (Exclude (Pac (PacNameAndVersion "Pac1" Nothing)))
+-- testJSON :: Spec
```
nevrome, line 331: Why did we have the ToJSON instance of `SignedEntity`? I guess you realized that we don't use it any more (?), removed it and then deactivated these tests here, right?

Maybe we should rewrite these tests to cover the show instance now? That would be a good use for this code. The `decode` tests can be removed, as we already have tests for `readEntitiesFromString` above.

test/PoseidonGoldenTests/GoldenTestsRunCommands.hs
```diff
@@ -848,6 +846,7 @@ testPipelineForge testDir checkFilePath = do
     runAndChecksumFiles checkFilePath testDir action17 "forge" [
           "forge" </> "ForgePac17" </> "ForgePac17.janno"
         ]
+    return ()
```
nevrome, line 849: Why is this return statement necessary here now?

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
```diff
@@ -1,3 +1,2 @@
 XXX018	F	POP3
-XXX099	M	POP3
```
nevrome, line 2: OK - why is XXX099 not in the output package any more? The selection string is `<XXX099>,<Lamnidis_2018-1.0.1:POP1:XXX017>,<Lamnidis_2018-1.0.0:POP3:XXX018>` and the data in `test/testDat/testPackages/ancient` did not change. I don't understand why it gets excluded.

nevrome (changes requested):
I did not think through the new tests in `EntitiesListTest`. Should I? I'm a bit worried about the behaviour of forge test #14.

stschiff:
Thanks for the comments and the detailed review. Some quick replies:

> That sounds like an all around very good solution! Your detailed analysis of the forgeScript semantics was very inspiring and a great learning experience for me. `indInfoConformsToEntitySpec` is very beautiful now.
> 
> That being said, here are some comments in no particular order:
> 
> * Removing the double-purpose of `IndividualInfo` sounds very sensible. I'm a bit confused why you decided right away to give it yet another double role as a replacement for `ExtendedIndividualInfo`, though? `indInfoAddCols` and the `infInfoIsLatest = False` hack are a pretty clear indication that this change does not come naturally, and that `IndividualInfo` is again shoehorned into two roles. I think it would be better to be consistent and keep `ExtendedIndividualInfo` as a separate type for the server-client communication. Maybe even with a more explicit name.

Yes, I also found this - from hindsight - unsatisfying. And I agree, we should bring back `ExtendedIndividualInfo`. One thing I'm grappling with is whether we want to also encode `isLatest` in there. I realised that we could even include this into the JSON returned from the server without breaking communication with previous versions, because those previous versions would just ignore an additional field. And while we're at it, I wonder whether we should include `isLatest` into our `HasNameAndVersion` typeclass and the `PacNameAndVersion` type. I'll think about this.
> * I originally hoped that we could make the version selection feature a non-breaking change, but I understand your reasoning as laid out here and in [Making Inclusion/Exclusion Semantics of Forgescript more consistent #269](https://github.com/poseidon-framework/poseidon-hs/issues/269). We will have to increase the major version number yet again.

Yeah, this is annoying. Perhaps we can increase the second version number, which according to Haskell's PVP should be the second major version number. To be discussed.

> * This breaking change requires a good deal of documentation, both for the trident changelog and for the website. Maybe we can split the workload a bit... We can certainly reuse some of the text you wrote for the issue.

Yes indeed. It's time to formerly specify forgescript.

> * When we're already breaking things: How difficult would it be to order the output according to the entities in the forgeScript now? With your massive simplifications (I'm jealous of your genius,  when I look into this part of the code, btw.) it seems like we would only have to make some adjustments to `resolveEntityIndices` and `indInfoConformsToEntitySpecs` to get this functionality. Just curious what you think.

I think we should have this feature! And I think it should be fairly easy to implement. Could be a 15-line function that just does that, and then some boilerplate for the CLI and interface stuff. Perhaps I can make a suggestion.

> 
> With these general comments out of the way I will now look into the code. I especially want to see if my understanding of the forgeScript semantics are reflected in the tests' behaviour.

stschiff:
> I'm also not a big fan of your excessive use of the list monad. It's elegant, but hard to read. I guess that comes down to personal preferences.

You heathen! Give me some time and I'll have you pulled over to the dark side!

Jokes aside, I think the only alternatives to list-monads are:
1) List-comprehensions, which are lovely, but only work for very small jobs inside the loop
2) a `map customFunc`, where `customFunc` is then defined as a `where`. The downside is the additional mental burden of having to process yet another function definition, plus any local variables would have to be passed into `customFunc` which is ugly. 

I should really write a blog post about this. List-Monads truly stand out in contrast to R and python and practically all other languages I know.

src/Poseidon/EntitiesList.hs
```diff
@@ -171,87 +129,48 @@ instance EntitySpec PoseidonEntity where
...
-                groupBy (\(_,IndividualInfo a _ _,_) (_,IndividualInfo b _ _,_) -> a == b)
-        resolveDuplicatesIfPossible :: [[(Int, IndividualInfo, [SelectionState])]] -> [[(Int, IndividualInfo, [SelectionState])]]
-        resolveDuplicatesIfPossible = map onlyKeepSpecifics
-        onlyKeepSpecifics :: [(Int, IndividualInfo, [SelectionState])] -> [(Int, IndividualInfo, [SelectionState])]
-        onlyKeepSpecifics groupOfInds =
-            let highPrio = filter (\(_,_,level) -> any isSpecified level) groupOfInds
-            in if length xs > 1 && length highPrio == 1
-               then highPrio
-               else groupOfInds
+determineNonExistentEntities entities indInfos = do -- for loop over entities
```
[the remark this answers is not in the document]
stschiff: Gladly. The code is:

```Haskell
determineNonExistentEntities entities indInfos = do -- for loop over entities
    entity <- underlyingEntity <$> entities
    let indices = resolveEntityIndices [entity] indInfos
    True <- return $ null indices
    return entity
```

Two things happen:
1) We loop over all entities. This is done via `entity <- underlyingEntity <$> entities`. What appears on the right side of the arrow is a list. On the left side is a single item of the list.
2) We _filter_ our main loop for cases in which `null indices` is True. Such a filter is also a feature of list-comprehensions, btw. This works because the list-monad has `fail` implemented as a skip in the loop, so a Pattern-match fail in a list-monad simply skips the respective loop-iteration.

src/Poseidon/CLI/Forge.hs
```diff
@@ -134,22 +135,22 @@ runForge (
...
+    -- determine indizes of relevant individuals
+    let relevantIndices = resolveEntityIndices relevantEntities allInds
+        duplicateReport = reportDuplicateIndividuals . map (allInds !!) $ relevantIndices
     -- check if there still are duplicates and if yes, then stop
-    unless (null unresolvedDuplicatedInds) $ do
+    unless (null duplicateReport) $ do
         logError "There are duplicated individuals, but forge does not allow that"
         logError "Please specify in your --forgeString or --forgeFile:"
-        mapM_ (\(_,i@(IndividualInfo n _ _),_) -> logError $ show (SimpleInd n) ++ " -> " ++ show (SpecificInd i)) $ concat unresolvedDuplicatedInds
+        mapM_ (\(IndividualInfo n _ _ _ _, sugg) -> logError $ show (Ind n) ++ " -> " ++ show sugg) duplicateReport
```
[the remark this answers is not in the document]
stschiff: OK, good point. I will add a test.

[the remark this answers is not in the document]
stschiff: Hmm, this isn't trivial, because it's log output, which we currently suppress in `testLog`. I guess I need to figure out how to funnel stderr into a file and then add that to the GoldenTests. But I guess that would be a good thing to have.

test/Poseidon/EntitiesListSpec.hs
```diff
@@ -3,16 +3,20 @@
...
+                                        SignedEntity (..))
 import           Poseidon.Package      (PackageReadOptions (..),
                                         PoseidonPackage (..),
                                         defaultPackageReadOptions,
                                         getJointIndividualInfo,
                                         readPoseidonPackageCollection)
 import           Poseidon.Utils        (PoseidonException, testLog)
 
-import           Data.Aeson            (decode, encode)
+-- import           Control.Monad (forM)
```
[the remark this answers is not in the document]
stschiff: Yes. Done

test/Poseidon/EntitiesListSpec.hs
stschiff: Can't exactly remember, but indeed I think I removed JSON instances where I didn't see use for. But you're right, we could test the Show instances.

stschiff: Done.

test/PoseidonGoldenTests/GoldenTestsRunCommands.hs
stschiff: ha, good catch. I guess I removed it to stop some compiler errors at some point, but clearly not necessary. I removed it.

src/Poseidon/CLI/Forge.hs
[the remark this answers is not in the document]
nevrome: I'm aware - yes. That's why I rather meant: "Did you run some tests on the command line?". Would be excellent to have this properly tested, but maybe it's not worth the effort.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
stschiff: Ha, well that one is quite tricky and points to a potential issue with our current semantics (easy to fix, though). So what's going on here is that `<XXX099>` is not in the _latest_ package, so it slips through the selection. This is the correct behavior given our exact semantics.

Here is the quote:
`<Ind1>: Select all individuals named "Ind1" in all latest packages.`

But I suppose we could change that behaviour to
`<Ind1>: Select the individual(s) named "Ind1" in the latest package where they are available`

src/Poseidon/EntitiesList.hs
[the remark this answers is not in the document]
stschiff: Will do!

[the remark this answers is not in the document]
stschiff: Done.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
stschiff: Actually no, I think our current behavior is correct. If someone takes out an individual in the latest version of a package, then by default we _should_ exclude it... unless the user specifically asks for it using the Specific-ind Syntax.

Also, note that in this case there would be a warning from `determineNonExistentEntites`, so the user would at least be confronted with the fact that this individual cannot be found (which might be surprising, but at least things aren't put under the carpet)

src/Poseidon/CLI/Forge.hs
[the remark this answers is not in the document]
stschiff: OK, so indeed I think getting the error output into the golden tests is possible but a bit of an overkill for now. I just ran a manual test, and it looks OK:

```
stephan_schiffels@biomac04 ancient % trident forge -f "*Lamnidis_2018-1.0.0*, *Lamnidis_2018-1.0.1*" -d . -o ~/Desktop/test
trident v1.3.0.4 for poseidon v2.5.0, v2.6.0, v2.7.0, v2.7.1
https://poseidon-framework.github.io

[Info]    5 found
[Info]    Checking Poseidon versions... 
[Warning] Duplicate individuals in package collection (XXX011, XXX012, XXX013, ...)
[Info]    Packages loaded: 5
[Info]    Unpackaged genotype data files loaded: 0
[Info]    Forging with the following entity-list: *Lamnidis_2018-1.0.0*, *Lamnidis_2018-1.0.1*
[Info]    2 packages contain data for this forging operation
[Error]   There are duplicated individuals, but forge does not allow that
[Error]   Please specify in your --forgeString or --forgeFile:
[Error]   <XXX011> -> <Lamnidis_2018-1.0.0:POP1:XXX011>
[Error]   <XXX011> -> <Lamnidis_2018-1.0.1:POP1:XXX011>
[Error]   <XXX012> -> <Lamnidis_2018-1.0.0:POP2:XXX012>
[Error]   <XXX012> -> <Lamnidis_2018-1.0.1:POP2:XXX012>
[Error]   <XXX013> -> <Lamnidis_2018-1.0.0:POP1:XXX013>
[Error]   <XXX013> -> <Lamnidis_2018-1.0.1:POP1:XXX013>
[Error]   <XXX014> -> <Lamnidis_2018-1.0.0:POP2:XXX014>
[Error]   <XXX014> -> <Lamnidis_2018-1.0.1:POP2:XXX014>
[Error]   <XXX015> -> <Lamnidis_2018-1.0.0:POP2:XXX015>
[Error]   <XXX015> -> <Lamnidis_2018-1.0.1:POP2:XXX015>
[Error]   <XXX016> -> <Lamnidis_2018-1.0.0:POP2:XXX016>
[Error]   <XXX016> -> <Lamnidis_2018-1.0.1:POP2:XXX016>
[Error]   <XXX017> -> <Lamnidis_2018-1.0.0:POP1:XXX017>
[Error]   <XXX017> -> <Lamnidis_2018-1.0.1:POP1:XXX017>
[Error]   <XXX018> -> <Lamnidis_2018-1.0.0:POP3:XXX018>
[Error]   <XXX018> -> <Lamnidis_2018-1.0.1:POP3:XXX018>
[Error]   <XXX019> -> <Lamnidis_2018-1.0.0:POP1:XXX019>
[Error]   <XXX019> -> <Lamnidis_2018-1.0.1:POP1:XXX019>
[Error]   Error in the forge selection: Unresolved duplicated individuals
```

Of course, one could try to be even smarter and kind of group per-individual. I'll take a quick look whether that makes sense.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
stschiff: OK, to be sure, Just tested this:

```
stephan_schiffels@biomac04 ancient % trident forge -f "<XXX099>" -d . -o ~/Desktop/test
trident v1.3.0.4 for poseidon v2.5.0, v2.6.0, v2.7.0, v2.7.1
https://poseidon-framework.github.io

[Info]    5 found
[Info]    Checking Poseidon versions... 
[Warning] Duplicate individuals in package collection (XXX011, XXX012, XXX013, ...)
[Info]    Packages loaded: 5
[Info]    Unpackaged genotype data files loaded: 0
[Info]    Forging with the following entity-list: <XXX099>
[Warning] Detected entities that do not exist in the dataset. They will be ignored: <XXX099>
[Info]    0 packages contain data for this forging operation
[Error]   Nothing to be forged
```

OK, so this is a bit confusing as it says `[Warning] Detected entities that do not exist in the dataset. They will be ignored: <XXX099>` which is not true. I will amend the error message.

stschiff: OK, so I've changed that warning. Now it's:

[Warning] The following entities could not be found in the dataset and will be ignored (keep in mind that without further specification, only latest versions of packages are considered): <XXX099>

src/Poseidon/CLI/Forge.hs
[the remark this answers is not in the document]
stschiff: OK, I have changed this now to the following:

```
[Error]   There are duplicated individuals, but forge does not allow that
[Error]   Please specify in your --forgeString or --forgeFile:
[Error]   Duplicate individual <XXX011> (please specify)
[Error]     <XXX011> -> <Lamnidis_2018-1.0.0:POP1:XXX011>
[Error]     <XXX011> -> <Lamnidis_2018-1.0.1:POP1:XXX011>
[Error]   Duplicate individual <XXX012> (please specify)
[Error]     <XXX012> -> <Lamnidis_2018-1.0.0:POP2:XXX012>
[Error]     <XXX012> -> <Lamnidis_2018-1.0.1:POP2:XXX012>
[Error]   Duplicate individual <XXX013> (please specify)
[Error]     <XXX013> -> <Lamnidis_2018-1.0.0:POP1:XXX013>
[Error]     <XXX013> -> <Lamnidis_2018-1.0.1:POP1:XXX013>
[Error]   Duplicate individual <XXX014> (please specify)
[Error]     <XXX014> -> <Lamnidis_2018-1.0.0:POP2:XXX014>
[Error]     <XXX014> -> <Lamnidis_2018-1.0.1:POP2:XXX014>
[Error]   Duplicate individual <XXX015> (please specify)
[Error]     <XXX015> -> <Lamnidis_2018-1.0.0:POP2:XXX015>
[Error]     <XXX015> -> <Lamnidis_2018-1.0.1:POP2:XXX015>
[Error]   Duplicate individual <XXX016> (please specify)
[Error]     <XXX016> -> <Lamnidis_2018-1.0.0:POP2:XXX016>
[Error]     <XXX016> -> <Lamnidis_2018-1.0.1:POP2:XXX016>
[Error]   Duplicate individual <XXX017> (please specify)
[Error]     <XXX017> -> <Lamnidis_2018-1.0.0:POP1:XXX017>
[Error]     <XXX017> -> <Lamnidis_2018-1.0.1:POP1:XXX017>
[Error]   Duplicate individual <XXX018> (please specify)
[Error]     <XXX018> -> <Lamnidis_2018-1.0.0:POP3:XXX018>
[Error]     <XXX018> -> <Lamnidis_2018-1.0.1:POP3:XXX018>
[Error]   Duplicate individual <XXX019> (please specify)
[Error]     <XXX019> -> <Lamnidis_2018-1.0.0:POP1:XXX019>
[Error]     <XXX019> -> <Lamnidis_2018-1.0.1:POP1:XXX019>
[Error]   Error in the forge selection: Unresolved duplicated individuals
```

... more verbose, but arguably more helpful!

stschiff:
OK, I have now addressed all comments. In particular, I have reintroduced `ExtendedIndividualInfo`. 

I have further changed the way `isLatest` is used. I have turned this now into a computed property, instead of storing it within `IndividualInfo`. It made me a bit uneasy that, for this to work, one could not set this property upon construction of a single `IndividualInfo`, but had to do it after the creation of a collection of IndividualInfos. This created an intermediate illegal state (with `isLatest` being falsely set to a dummy `False` or `True`). I have therefore gone back to _not_ storing `isLatest` anywhere, but computing it on the fly. I could not detect any performance problems thus far. And it's arguably more secure now.

Your comments above have all been addressed. There are two open conversations that you can check and resolve if you agree, or further discussed.

src/Poseidon/CLI/Forge.hs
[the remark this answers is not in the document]
nevrome: Beautiful :+1:

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
nevrome: I think I would prefer a tall format for this message, when we're already at it. Happy to add a comment with the following suggestion:

```
[Warning] The following entities could not be found in the dataset and will be ignored
[Warning] <XXX099>
[Warning] <XXX105>
[Warning] Maybe these entities exist in older package versions?
```

What's your opinion?

nevrome:
I wonder if we should directly introduce an `--onlyLatest` argument to `list`. I think I would want to filter to the set of latest packages for most queries.

This problem also extents to `qjanno` and the `janno` R package. Always existed, but now that we embrace multi-version  repositories, it becomes a real issue. Both tools naively crawl for `*.janno` files, which can lead to very wrong results. I will open issues in their repositories to discuss this.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
stschiff: Love it. Will make that change and post here when done.

stschiff:
> Alright, alright - I'll try to look at the list monad with a new admiration.
> 
> Measuring `isLatest` when it's needed is conceptually cleaner. And as long as it's not too expensive we can do it like this.
> 
> I think the webserver does not return an `isLatest` field in the package, group and individual APIs, right? This may be an important addition (see e.g. [poseidon-framework/poseidon-framework.github.io#45 (comment)](https://github.com/poseidon-framework/poseidon-framework.github.io/pull/45#issuecomment-1710737885)). I also think `list` should return this information in an extra column. I can make a suggestion for these changes. Or did you decide not to implement this on purpose?

Why would you want the web server to return this, when we can also compute it client-side after we download the data? Do you mean because somebody might use the Web-API outside of trident and therefore would appreciate this info? Maybe... hmm. perhaps good to discuss further in the issue you've opened.

stschiff:
> I wonder if we should directly introduce an `--onlyLatest` argument to `list`. I think I would want to filter to the set of latest packages for most queries.
> 
> This problem also extents to `qjanno` and the `janno` R package. Always existed, but now that we embrace multi-version repositories, it becomes a real issue. Both tools naively crawl for `*.janno` files, which can lead to very wrong results. I will open issues in their repositories to discuss this.

Yes. I love the `--onlyLatest` option for `trident list`. Let's discuss it further in the other issue.

nevrome:
> > I think the webserver does not return an `isLatest` field in the package, group and individual APIs, right? This may be an important addition (see e.g. [poseidon-framework/poseidon-framework.github.io#45 (comment)](https://github.com/poseidon-framework/poseidon-framework.github.io/pull/45#issuecomment-1710737885)). I also think `list` should return this information in an extra column. I can make a suggestion for these changes. Or did you decide not to implement this on purpose?
> 
> Why would you want the web server to return this, when we can also compute it client-side after we download the data? Do you mean because somebody might use the Web-API outside of trident and therefore would appreciate this info?

Yes - that was my idea. Otherwise every client software has to write the same `map head . group_by . sort` operation, right? For example for our website we would need this code in Javascript. Not super difficult to write, but the server could provide this info right away for a little more convenience...

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
stschiff: OK, I've made this change now. By the way, why have we not made this an actual error? Shouldn't we fail if the user looks for individuals that cannot be found? That would make sense to me!

stschiff:
> > Yes, I also found this - from hindsight - unsatisfying. And I agree, we should bring back ExtendedIndividualInfo. One thing I'm grappling with is whether we want to also encode isLatest in there. I realised that we could even include this into the JSON returned from the server without breaking communication with previous versions, because those previous versions would just ignore an additional field. And while we're at it, I wonder whether we should include isLatest into our HasNameAndVersion typeclass and the PacNameAndVersion type. I'll think about this.
> 
> Adding the `isLatest` information to `ExtendedIndividualInfo` and therefore to the Web-API is a good idea, in my opinion. And I would rather have it in an extra field and not appended to the version number, if this is possible. Just to make parsing easier.
> 
> > Yeah, this is annoying. Perhaps we can increase the second version number, which according to Haskell's PVP should be the second major version number. To be discussed.
> 
> Yes - exactly - that's what I meant. Sorry for not making this clear. When I speak about the major version number, then I mean the second one in PVP. The first one may be called the era number? I'm not sure.
> 
> > I think we should have this feature! And I think it should be fairly easy to implement. Could be a 15-line function that just does that, and then some boilerplate for the CLI and interface stuff. Perhaps I can make a suggestion.
> 
> That would be so neat!

I decided to not implement this within this PR. But I've made a reminder into #230

stschiff:
> > I wonder if we should directly introduce an `--onlyLatest` argument to `list`. I think I would want to filter to the set of latest packages for most queries.
> > This problem also extents to `qjanno` and the `janno` R package. Always existed, but now that we embrace multi-version repositories, it becomes a real issue. Both tools naively crawl for `*.janno` files, which can lead to very wrong results. I will open issues in their repositories to discuss this.
> 
> Yes. I love the `--onlyLatest` option for `trident list`. Let's discuss it further in the other issue.

OK, I have implemented this feature now. Also updated the golden tests.

stschiff:
From my point of view, this is now ready. There was one open conversation which I have resolved, and added one new issue for a side-discussion. Otherwise, from my perspective this is good to go.

I have now changed the base of this PR, so this would be merged into master and thereby also close #260. 

Let me know what you think, I can then do the merge and adapt the documentation

stschiff:
Sorry, I've put this back in draft mode, as I've noticed some points in Xerxes that might lead to small changes here. Will set to green once those changes are in.

stschiff:
OK, this is ready for review now. I refactored some more bits to make them re-usable from within xerxes.
Also, very importantly, I've added a genotype-ploidy check. Because Genotype Ploidy is now used in xerxes, we check for consistency now.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac14/ForgePac14.ind
nevrome: Yes - probably a good idea to make it an error. I see that you already implemented this now. This means this conversation is resolved, finally :+1:

nevrome:
Excellent - some last comments (I hope):

- [ ] I think `list` should have an extra, boolean column for the `latest` property. A `(latest)` suffix to the version number is harder to parse, especially for quick analyses on the command line.
- [ ] A number of subcommands (`genoconvert`, `rectify`, `survey`, `summarize`, `validate`) now only operate on the latest package versions in a given base directory. This may be confusing -- especially if there is no clear need to only consider one iteration of a package (e.g. in `genoconvert`, `survey` or `validate`). Should we make them multi-version compatible? Maybe they should behave like `list`: Always consider all package versions, except the `--onlyLatest` flag is set. This doesn't fully match what users might expect for `rectify` and `summarize`, but it would be a universal and consistent solution. We could essentially set `_readOptKeepMultipleVersions` in `defaultPackageReadOptions` to true.
- [ ] The technical, internal changelog should probably also mention the high-level refactoring within the library: The `SecondaryTypes` module was dissolved and instead we now have `Contributor`, `ServerClient` and `Version`. `EntitiesList` (or what is left of it) now lives in `EntityTypes`.
- [ ] The way we handle versions now (e.g. in `parseVersion`) is a bit imprecise. According to the implementation we only accept versions of the form `X.Y.Z`, but the Poseidon v2.7.1 standard does not specifically say this. I think we should make this constraint explicit in the next version of the standard.

Let me know if you agree with these ideas and whether I should take over and implement some of them.

stschiff:
OK, list now has its own column for "Is latest". I also noticed that the raw output so far did not print out the column headings, which is however vital, in particular with additional janno columns when listing individuals, so I now added this to the raw output (see for example golden test output for test-pipeline `list4`.

stschiff:
Regarding the second point, I have now changed the default behaviour. I looked at the golden test outputs, and think that `summarise` looks strange when applied to multiple versions. I think it makes no sense in that specific case, so I explicitly now set `_readOptKeepMultipleVersions = False` in `summarise`. Happy to discuss

stschiff:
Regarding points 3 and 4 I agree! I think you wanted to take a stab at the release changelog, and regarding point 4 yes, I will make a PR to the schema (no Poseidon version change, just a clarification, I think)

src/Poseidon/Package.hs
```diff
@@ -285,6 +285,17 @@ readPoseidonPackageCollection opts baseDirs = do
             if _readOptOnlyLatest opts
             then filterM (isLatestInCollection loadedPackages) loadedPackages
             else pure $ loadedPackages
+    -- check if any packages appear more than once
```
stschiff, line 289: Why did you introduce this check? I mean, I get it, you want to avoid having _exactly_ the same package (and version) present, but it seems a rather unlikely case for me to cater for. I mean... why not, but I don't think it's related to the new changes we discussed last, right?

src/Poseidon/Package.hs
```diff
@@ -809,41 +769,72 @@ writePoseidonPackage (PoseidonPackage baseDir ver tit des con pacVer mod_ geno j
...
-            colNames -> [(k, BSC.unpack <$> toNamedRecord jannoRow HM.!? BSC.pack k) | k <- colNames]
-    return $ ExtendedIndividualInfo name groups pacName version additionalColumnEntries
+    return $ IndividualInfo
+        (jPoseidonID jannoRow)
+        ((getJannoList . jGroupName) jannoRow)
+        (makePacNameAndVersion pac)
 
+getExtendedIndividualInfo :: (MonadThrow m) => [PoseidonPackage] -> [String] -> m [ExtendedIndividualInfo]
+getExtendedIndividualInfo allPackages additionalJannoColumns = do
+    concatForM allPackages $ \pac -> do
```
stschiff, line 811: This definitely is much easier and shorter with sequence and list-monad. I want to make the change.

stschiff: OK, I've made the change. Check out my last commit, I hope you agree.

stschiff:
OK very cool, I've reviewed the code changes and ran a few command line tests on the test-data. I'm happy. What do you say? I say, we merge and start a test-release, then get the rest of the ball going. Of course, finalist the release only after further tests and update of the server.

stschiff:
So, you've decided to stick with my suggestion and _not_ include an `--onlyLatest` option for `summaries`? I might have second thoughts...

stschiff:
I've added a fix for validate, which would fail if some packages were skipped if `--onlyLatest` is selected.

src/Poseidon/Package.hs
nevrome: Well - it's an important constraint, because loading copies of a package can have unforeseen consequences. And I can easily imagine a situation where a user accidentally copies a package and then runs into this situation.

And I think this is just a more general implementation of the rule you suggested yesterday for `isLatestInCollection`...

src/Poseidon/Package.hs
nevrome: Fascinating!

src/Poseidon/Package.hs
stschiff: OK I agree

PR #275, opened 2023-09-26
I turned off the ploidy check

nevrome:
As discussed. I will open an issue now and document the problem for future intervention.


stschiff (approved):
OK, have you tried this out on the server, restarted and made sure that all packages are read properly? That would be the necessary check for me, then it'd be good to go!

PR #280, opened 2023-10-26
V 1.4.0.3: Optimization of resolveEntityIndices

stschiff:
As already somewhat touched upon around the code changes introduced in 1.4.0.0, we had a severe performance leak in `resolveEntityIndices`, which affected fetch, forge, and crept into xerxes as well.

I solved this now by simply computing the isLatest vector only once when I create individualInfos from packages (`getJointIndividualInfo`). For that purpose I have introduced a new type synonym for the tuple of indInfos and Bools, defined in `EntityTypes`:

```
type IndividualInfoCollection = ([IndividualInfo], [Bool]) 
```

Key downstream functions `determineRelevantPackages`, `resolveEntityIndices`, `resolveUniqueEntityIndices`, `determineNonExistentEntities` and `checkIfAllEntitiesExist` all now use this new tuple, which saves them from calling `isLatestInCollection` themselves. 

As far as I can see, this solved the performance bug completely. 

This command line is diagnostic:

```
trident fetch -d comp_book_fstats_data/ -f "*2010_RasmussenNature*"
```

Before this change, this took ages before actually starting to download. Now it starts almost immediately. 

You can test from this branch via

```
stack run trident -- fetch -d comp_book_fstats_data/ -f "*2010_RasmussenNature*"
```

I've bumped the version number and added a Changeling. I will now update xerxes as well and will see whether there is anything in the new API that needs further change.

### DIFF
--- a/.github/workflows/Dockerfile.centos
+++ b/.github/workflows/Dockerfile.centos
@@ -4,9 +4,9 @@ RUN yum -y update
 RUN yum -y install zlib-devel wget ncurses-devel ncurses-compat-libs make gcc
 
 # Install GHC since stack's local install has issues
-RUN wget https://downloads.haskell.org/~ghc/9.2.7/ghc-9.2.7-x86_64-centos7-linux.tar.xz
-RUN tar xvf ghc-9.2.7-x86_64-centos7-linux.tar.xz
-RUN cd ghc-9.2.7; ./configure; make install
+RUN wget https://downloads.haskell.org/~ghc/9.4.7/ghc-9.4.7-x86_64-centos7-linux.tar.xz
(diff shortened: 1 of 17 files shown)


stschiff:
OK, I've updated now to LTS-21.17. Build and tests work without `--pedantic`, but one warning breaks pedantic compilation:
```
Survey.hs:98:66: warning: [-Wtype-equality-requires-operators]
    The use of ‘~’ without TypeOperators
    will become an error in a future GHC release.
    Suggested fix: Perhaps you intended to use TypeOperators
   |                         
98 |         getRatiosForEachField :: (Generics.SOP.Generic a, Code a ~ '[ xs ], All PresenceCountable xs) => [a] -> [Ratio Int]
   |                                                                  ^
```

nevrome (approved):

Issue #276, opened 2023-09-26
Breaking changes and maintaining reproducibility


nevrome:
A recent change added in trident v1.4.0.0 (#268) and again reverted in v.1.4.0.2 (#275) uncovered an interesting issue with our current package version management concept for Poseidon:

This change boils down to a new cross-file consistency check: The genotype ploidy (`Genotype_Ploidy`) documented in .janno files was supposed to be checked when reading a package, and if a sample marked as `haploid` had a heterozygote SNP, this would throw an error. So this change in trident did not reflect a change in the Poseidon schema, but only a more strict validation of an already formalized promise in the schema.

Activating this check had unforeseen consequences, because some packages in our public community archive actually failed this validation step and had to be fixed (https://github.com/poseidon-framework/community-archive/pull/135). They were, strictly speaking, not valid Poseidon packages before this update. With the new, adjusted versions of said packages, we considered this issue solved.

This was premature, though, because trident v1.4.0.0 and v1.4.0.1 were not able to read the old package versions any more. For a user this is not necessarily a problem: They can either switch to the new, updated package versions, or continue working with an old trident release, which doesn't check this property. But a big problem emerged for the Poseidon web server (`trident serve`): It could neither read nor serve the old package versions any more, thus breaking the reproducibility promise that motivated the entire shift to a versioned web API.

We averted chaos by deactivating the ploidy check in  v.1.4.0.2. But this is only a temporary solution, because we want to include this check eventually. A similar issue would probably also emerge, if we ever decided to update the Poseidon schema in a breaking way.

There are various ways to solve this issue permanently. Here are two potential solutions (formulated in a way that clearly highlights which one I prefer :grimacing:):

1. Loosen the requirements on Poseidon packages in `trident`'s reading process to a point where even broken or anachronistic packages could be loaded. Issues could be pointed out as warnings. `validate` could continue to be strict. This way the server could easily read and serve every somewhat functional package and even outdated versions. Users may experience less issues when using trident, but may end up with potentially broken output (as a consequence of broken input). This explicitly includes packages that do not fulfill the guarantees of the Poseidon schema.
[...]

PR #283, opened 2023-12-08
More informative consistency checks

nevrome:
... and other ideas to make modifying .janno files less painful with better validation.

Work in progress. Not sure if all of these ideas are useful.

### DIFF
--- a/src/Poseidon/Janno.hs
+++ b/src/Poseidon/Janno.hs
@@ -998,7 +998,7 @@ readJannoFileRow :: FilePath -> (Int, Bch.ByteString) -> PoseidonIO (Either Pose
 readJannoFileRow jannoPath (lineNumber, row) = do
     case Csv.decodeByNameWith decodingOptions row of
         Left e -> do
-            return $ Left $ PoseidonFileRowException jannoPath lineNumber $ removeUselessSuffix e
+            return $ Left $ PoseidonFileRowException jannoPath (show lineNumber) $ removeUselessSuffix e
         Right (_, jannoRow :: V.Vector JannoRow) -> do
             case checkJannoRowConsistency jannoPath lineNumber $ V.head jannoRow of
                 Left e -> do
@@ -1041,7 +1041,8 @@ checkJannoRowConsistency :: FilePath -> Int -> JannoRow -> Either PoseidonExcept
 checkJannoRowConsistency jannoPath row x =
     case check of
         Right res -> Right res
-        Left  e   -> Left $ PoseidonFileRowException jannoPath row e
+        Left  e   -> Left $
+            PoseidonFileRowException jannoPath (show row ++ " (" ++ jPoseidonID x ++ ")") e
     where
         check :: Either String JannoRow
         check =
--- a/src/Poseidon/Janno.hs
+++ b/src/Poseidon/Janno.hs
@@ -1067,16 +1067,16 @@ allEqual x  = length (nub x) == 1
 
 checkC14ColsConsistent :: JannoRow -> Either String JannoRow
 checkC14ColsConsistent x =
-    let isTypeC14       = jDateType x == Just C14
-        lLabnr          = getCellLength $ jDateC14Labnr x
-        lUncalBP        = getCellLength $ jDateC14UncalBP x
-        lUncalBPErr     = getCellLength $ jDateC14UncalBPErr x
-        anyColFilled    = lLabnr > 0 || lUncalBP > 0 || lUncalBPErr > 0
-        anyMainColEmpty = lUncalBP == 0 || lUncalBPErr == 0
-        allSameLength   = allEqual [lLabnr, lUncalBP, lUncalBPErr] ||
+    let isTypeC14        = jDateType x == Just C14
+        lLabnr           = getCellLength $ jDateC14Labnr x
+        lUncalBP         = getCellLength $ jDateC14UncalBP x
+        lUncalBPErr      = getCellLength $ jDateC14UncalBPErr x
+        anyMainColFilled = lUncalBP > 0 || lUncalBPErr > 0
+        anyMainColEmpty  = lUncalBP == 0 || lUncalBPErr == 0
+        allSameLength    = allEqual [lLabnr, lUncalBP, lUncalBPErr] ||
                           (lLabnr == 0 && lUncalBP == lUncalBPErr)
-    in case (isTypeC14, anyColFilled, anyMainColEmpty, allSameLength) of
-        (False, True, _, _ )    -> Left "Date_Type is not \"C14\", but either Date_C14_Labnr, \
+    in case (isTypeC14, anyMainColFilled, anyMainColEmpty, allSameLength) of
+        (False, True, _, _ )    -> Left "Date_Type is not \"C14\", but either \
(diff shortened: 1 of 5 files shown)


nevrome:
As indicated above this PR introduces a number of changes to improve the error reporting for broken .janno files. @AyGhal recently contacted me about various error messages she encountered when preparing packages for the PCA. Some of them were not terribly useful. So here's what I did to mitigate this issue:

1. When cassava encounters a parsing issue in an individual .janno file field it returns an error. For many more specific fields with custom types we created smart constructors with explicit error reporting (called by the respective `parseField` functions). For them the error messages are good and useful. But for types without this additional specification, e.g. simple `Int` or `Double` fields, the messages are very simple, e.g.:

    ```
    parse error (Failed reading: conversion error: expected Int, got "430;" (incomplete field parse, leftover: [59]))
    ```

    Especially the list of ASCII codes for the unparsed leftover is disappointing. cassava unfortunately does not expose the components of this message in a useful error type. It only returns a string. I thus went ahead and parsed this string to compile something slightly more readable:

    ```
    parse error in one column (expected data type: Int, broken value: "430;", problematic characters: ";")
    ```

    A pretty disgusting solution and incomplete at best (I don't see how to get the column name into this message), but a bit better than what we had so far.

2. A problem that was raised multiple times by @AyGhal was the exact behaviour of the consistency check for radiocarbon dates in `checkC14ColsConsistent`. I reworked that now into a tree of allowed and forbidden cases, whose leaves feature clear error messages (beyond the pretty useless `The Date_* columns are not consistent`). For example if the the `Date_Type` is not set to `C14`, but we have radiocarbon date information in the data columns, we now get this:

    ```
    Date_Type is not "C14", but either Date_C14_Uncal_BP or Date_C14_Uncal_BP_Err are not empty.
    ```

    The check's logic was changed in this process and I would like to ask you to take a close look into it in the review. I implemented the same system for `Contamination_*` and the `Relation_*` columns, but the validation logic is much simpler there.
3. The most interesting/controversial change, finally, is the way these consistency checks are called. So far they have been part of an `Either` based system that could only fail or forward the valid .janno row. `checkJannoRowConsistency` and the functions below it now live in a little monad stack of `Except` and `Writer`, which allows to record both failure and (!) warnings.

    ```
    type JannoRowLog = E.ExceptT String (W.Writer JannoRowWarnings)
   ```

   I created this, because I encountered a critical omission in `checkC14ColsConsistent`, which allowed the `Date_Type` to be set to `C14` even if `Date_C14_Uncal_BP` and `Date_C14_Uncal_BP_Err` are empty. Two packages in the PCA (`2022_Gretzinger_AngloSaxons` and `2020_Nagele_Caribbean`) feature this invalid case and this brings us exactly in the kind of trouble discussed in #276. Through the new monad stack I now report only a warning for these packages instead of refusing to load them, thus propagating solution 1. mentioned there. What do you think?

I also quickly checked `SequencingSource.hs` and removed some unused code there to keep the two modules for .janno and .ssf files somewhat consistent.

stschiff:
Wow, OK, I will need some time to review this. Happy to discuss priorities, but I can see how this will be a useful addition!

nevrome:
@stschiff Just a quick reminder that I'm waiting for a review here :+1:

stschiff:
On it! Likely not finish today though.

src/Poseidon/Janno.hs
```diff
@@ -41,19 +41,23 @@ module Poseidon.Janno (
...
                                                        PoseidonIO, logDebug,
-                                                       logError,
+                                                       logError, logWarning,
                                                        renderPoseidonException)
 
 
 import           Control.Applicative                  (empty)
 import           Control.Exception                    (throwIO)
 import           Control.Monad                        (unless, when)
+import qualified Control.Monad.Except                 as E
```
stschiff, line 58: I think `Control.Monad.Except` and `Control.Monad.Writer` are from the `mtl` library, right? I think I prefered the `transformers` library in the past, and we might want to check in the future whether we can try and stick to one of them. But OK for now.

src/Poseidon/Janno.hs
```diff
@@ -1037,23 +1082,27 @@ checkJannoConsistency jannoPath xs
...
-          "The Contamination_* columns are not consistent"
-    | not $ checkRelationColsConsistent x = Left $ PoseidonFileRowException jannoPath row
-          "The Relation_* columns are not consistent"
-    | otherwise = Right x
-
-checkMandatoryStringNotEmpty :: JannoRow -> Bool
+-- Row-wise janno consistency checks
+
+type JannoRowWarnings = [String]
+type JannoRowLog = E.ExceptT String (W.Writer JannoRowWarnings)
```
stschiff, line 1088: As discussed, I think I would generally prefer the policy "If you want to log, you need PoseidonIO". Inventing a new Monad-Stack that preserves purity in such a narrow area is a bit of an overkill in my view. I think a Writer would make sense if we were to collect log-messages over much longer, transcending rows and even files, but for a single row? Anyway, as discussed, I have understood the code and I'm OK with it. Also, this kind of code might be useful in the future, so perhaps good to have a template that already works.

src/Poseidon/Janno.hs
```diff
@@ -1062,33 +1111,50 @@ allEqual :: Eq a => [a] -> Bool
...
+        anyMainColFilled = lUncalBP > 0 || lUncalBPErr > 0
+        anyMainColEmpty  = lUncalBP == 0 || lUncalBPErr == 0
+        allSameLength    = allEqual [lLabnr, lUncalBP, lUncalBPErr] ||
+                          (lLabnr == 0 && lUncalBP == lUncalBPErr)
+    in case (isTypeC14, anyMainColFilled, anyMainColEmpty, allSameLength) of
+        (False, False, _, _ )   -> return x
+        (False, True, _, _ )    -> E.throwError "Date_Type is not \"C14\", but either \
+                                         \Date_C14_Uncal_BP or Date_C14_Uncal_BP_Err are not empty"
+        (True, _, False, False) -> E.throwError "Date_C14_Labnr, Date_C14_Uncal_BP and Date_C14_Uncal_BP_Err \
+                                         \do not have the same lengths. Date_C14_Labnr can be empty"
```
stschiff, line 1129: I think the addition "Date_C14_Labnr can be empty" is not helping here. I would remove it. The error concerns the non-equal length, and the error should focus only on that. The positive message coming after muddles the main message, in my view. Otherwise you would also have to write it in the last case below, I think?

src/Poseidon/Janno.hs
nevrome: At the moment we **only** depend on `mtl` and not `transformers` in poseidon-hs. I didn't change any dependencies for this PR.

src/Poseidon/Janno.hs
nevrome: Your point is well taken :+1:

src/Poseidon/Janno.hs
nevrome: Removed.

src/Poseidon/Janno.hs
nevrome: I'm actually pretty indifferent here, because I don't know how `transformers` works. I heard people are using "effect-systems" instead of monad stacks nowadays. I'm not there yet... :grimacing:

PR #288, opened 2024-02-21
more adjustements for jannocoalesce

nevrome:
I did essentially three things in this PR:

1. Added an option to exclude columns. I did so with a custom record type, some logic to handle it and a new command line input setup. The user can now either select `--includeColumns` or `--excludeColumns`. If they choose neither, then `AllJannoColumns` applies.

```Haskell
data CoalesceJannoColumnSpec =
      AllJannoColumns
    | IncludeJannoColumns [BSC.ByteString]
    | ExcludeJannoColumns [BSC.ByteString]
 ```

2. Added some helpful logging about the number of fields changed and the number of target-source mismatches. This is particularly useful for large packages. Inspired by our recent discussion I solved this with `IORef` and not with a monad transformer. But this now also feels a bit verbose - not sure.

3. Found and fixed a bug in the handling of additional, not specified variables in the source .janno file. They would always get copied, even if `--fillColumns` (now `--includeColumns`) was set. I added some tests to cover this behavior.

    My solution was to rewrite the core logic of `mergeRow`. Instead of calculating a conditional union it now first creates a version of the target with exactly the desired values and then loops through them and replaces the right ones from the source. This solution seems to work and passes all tests. Maybe it's less efficient in some cases.


stschiff (approved):
OK great! I added a single comment to merge row to explain the dummy-fill-in. I think this is all a very elegant solution. Yes, the IORefs feel dirty somewhat, but given that we are already in impure code due to logging, I think it's OK.

nevrome:
Thanks - merged! I will now add a release-changelog on the jannocoalesce branch and then merge into master to publish there.

Issue #289, opened 2024-02-22
Option to remove contributors from a package with `trident rectify`, or automatic overwriting of Josiah Carberry when new contributors are added.


TCLamnidis:
Creating a new package semi-automatically is currently hampered by Josiah Carberry. in the contributor field.
I would like a way to create/update the contributor field to remove Josiah (and potentially, though less importantly, other contributors) from the list.

I would suggest either a `--removeContributor` option to `trident rectify`, to contrast the existing `--newContributor` option,
or a special case being added for Josiah Carberry, where he is automatically removed from the contributor list when another contributor is added to the package.

This would minimise the amount of manual editing of the POSEIDON.yml that a user has to do to finalise a package. That said, I can see that this is only an issue in cases where one user is creating many packages at a time, so maybe not a priority.

stschiff:
We already have a `--minimal` flag. How about we make it so that when that is chosen, no dummy values are filled in the YAML file?

TCLamnidis:
`--minimal` also creates no `janno`, or `bib` file, just a `POSEIDON.yml` to go with the genotypes.

stschiff:
Hmm, I'm concerned that we should think hard before adding more CLI options here for very specific YAML changes. The usual rectify-workflow is a quite specific edit of the YAML, geared towards "typical" update workflows. If you want custom changes to the YAML, you could just do that directly. For example, there are tools like https://github.com/andywerner/yaml_cli that you can check, which allow you to edit YAML from the CLI.

Or in fact you could do this:

```
python3 -c 'import yaml; d=yaml.load(open("POSEIDON.yml", "r").read(), Loader=yaml.CLoader); d["contributor"] = {"name": "Thiseas", "email" : "Thiseas.mail.com"};open("POSEIDON2.yml", "w").write(yaml.dump(d))'
```

TCLamnidis:
I would argue that removing Josiah Carberry from the contributors is very much a typical update workflow.
One could even argue that validation of a package for addition to the PCA _SHOULD_ check for his absence from the contributors.
[...]

PR #293, opened 2024-04-02
Retiring Josiah Carberry

nevrome:
This solves #289 if merged. I applied the following changes:

- removed Josiah from `newPackageTemplate`, so that he doesn't get added any more to new packages created by `init` and `forge` - the contributor field is missing in the output of these commands now
- adjusted the golden test output accordingly, but also added Josiah to one of the test packages (`test/testDat/testPackages/ancient/Schiffels_2016`) to keep him around at least in one way and make sure that the ORCID parser runs in the tests
- added a new warning to the package reading process to point out an empty contributor field:

```
[Warning] Contributor missing in POSEIDON.yml file of package 2010_RasmussenNature-2.1.1
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,3 @@
-# taken from https://vrom911.github.io/blog/github-actions-releases
-
 name: Release
 
 on:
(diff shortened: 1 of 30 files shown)


nevrome:
If we go with this, then we probably have to increase the major version number. Removing Josiah could break workflows.

stschiff (approved):
Yes, this certainly is a simple enough change to approve. You haven't yet touched version and Changelog. As discussed, from my point OK to increase the second PVP version nr for this - technically - breaking change. Thanks!

Issue #300, opened 2024-05-29
Arranging individuals in a poseidon package


TCLamnidis:
A feature in trident where it can arrange the individuals in a poseidon package by the Poseidon_ID or Group_ID would be nice to have. 

Practically, the operation is somewhat superfluous, as most tools do not care about the order of individuals on input. However, being able to rearrange the janno file in some basic ways without breaking the package can be very helpful for basic categorisation of information. Currently, changing the order of individuals in the janno is prohibited without also changing the order in the fam/ind file, which in turn means also changing the order in the bed/geno file.

This is a problem when dealing with reviewing PRs that add or update packages where the information is not in an intuitive order. Minotaur suffers from this problem particularly, because nextflow's asynchronous nature means that the order of individuals in the package will be different with each rerun of a package, and will always be random unless steps are taken to reorder janno/ind/geno files during package creation. 

The minotaur-specific issue of order when rerunning packages is somewhat mitigated by `jannocoalesce`, but reordering of individuals is a data manipulation that might be useful for many users, and it makes sense to do within trident since it requires fiddling with the actual genotype data as well as the janno. 

I assume this is technically already possible through some complicated `forge` command, where a user specifies all the poseidon_IDs in `<>` and in the desired order, but having an option or subcommand to sort the package without needing to explicitly state the order would be a nice quality of life feature imo.

nevrome:
> I assume this is technically already possible through some complicated forge command, where a user specifies all the poseidon_IDs in <> and in the desired order

If I'm not mistaken this is not the case. The order in forgeScript does not translate to the output order. It's a selection language only. The sample order in the `forge` output instead depends on the order in the source packages and the order by which these packages are discovered in the file system via `-d`.  The `forge` algorithm just checks for each input individual if they conform to the entity specification in the forgeScript, but does not change their order from the input sequence.

We discussed the idea to change this behaviour in the past in #230, but it would be a big change. @stschiff wrote back then:
> In full generality, this would require a complete rewrite of how we currently read the forge DSL.

Back then we decided to postpone this.

Now regarding a pure `arrange` or `sort` subcommand my gut feeling is that this would be more easy to realize :thinking:. We could allow to sort by a list of .janno columns. Entirely custom ordering would then be possible by adding an own column with the desired ranks.

TCLamnidis:
That would cover my use case perfectly, and then some!

stschiff:
[...]

PR #301, opened 2024-05-31
Ordered forge

stschiff:
Added an option to forge to output ordered according to the entities given.


stschiff:
This will partly address #300, and close #230

src/Poseidon/EntityTypes.hs
```diff
@@ -263,15 +264,35 @@ reportDuplicateIndividuals individuals = do -- loop over duplication groups
...
+resolveEntityIndicesUnordered entities (indInfos, areLatest) =
+    [ i | (i, ind, l) <- zip3 [0..] indInfos areLatest, indInfoConformsToEntitySpecs ind l entities ]
+
+-- | this finds the indices of all individuals from an individual-list which are specified in the Entity list, ordered by the entity list
+resolveEntityIndicesOrdered :: (EntitySpec a) => [a] -> IndividualInfoCollection -> [Int]
+resolveEntityIndicesOrdered entities (indInfos, areLatest) = go [] entities
+  where
+    go :: (EntitySpec a) => [Int] -> [a] -> [Int]
+    go selectedIndices [] = selectedIndices
+    go selectedIndices (entity:restEntities)=
```
nevrome, line 282: Neat! I think this works. But I wonder whether things would become slightly more easy to read by handling the two paths separately:

```haskell
go :: (EntitySpec a) => [Int] -> [a] -> [Int]
go selectedIndices (entity:restEntities)
   | (isPoseidonEntity entity || isIncludeEntity) = ...
   | isExcludeEntity entity = ...
```

Maybe not. If we only had `SignedEntity` to consider, then it would be a simple pattern match, but with the `EntitySpec` class we need a case statement or guards.

test/PoseidonGoldenTests/GoldenTestsRunCommands.hs
```diff
@@ -876,11 +892,31 @@ testPipelineForge testDir checkFilePath = do
...
+        , _forgeOutPacPath   = testDir </> "forge" </> "ForgePac18"
+        , _forgeOutPacName   = Just "ForgePac18"
+        , _forgePackageWise    = False
+        , _forgeOutputPlinkPopMode = PlinkPopNameAsFamily
+        , _forgeOutputOrdered = True
+    }
+    let action18 = testLog (runForge forgeOpts18) >> patchLastModified testDir ("forge" </> "ForgePac18" </> "POSEIDON.yml")
+    runAndChecksumFiles checkFilePath testDir action18 "forge" [
+          "forge" </> "ForgePac18" </> "ForgePac18.janno",
+          "forge" </> "ForgePac18" </> "ForgePac18.fam"
```
nevrome, line 918: Maybe the .bed file as well as its order also depends on on `_forgeOutputOrdered = True`.

src/Poseidon/EntityTypes.hs
```diff
@@ -285,13 +306,13 @@ resolveUniqueEntityIndices entities indInfoCollection = do
     return relevantIndices
 
 -- | this returns a list of entities which could not be found
-determineNonExistentEntities :: (MonadThrow m, EntitySpec a) => [a] -> IndividualInfoCollection -> m EntitiesList
-determineNonExistentEntities entities indInfoCollection = do
-    return [ entity | entity <- map underlyingEntity entities, indices <- resolveEntityIndices [entity] indInfoCollection, null indices]
+determineNonExistentEntities :: (EntitySpec a) => [a] -> IndividualInfoCollection -> EntitiesList
+determineNonExistentEntities entities indInfoCollection =
+    [ entity | entity <- map underlyingEntity entities, null (resolveEntityIndices False [entity] indInfoCollection)]
```
nevrome, line 311: I think 

```haskell
[... | ..., indices <- resolveEntityIndices [entity] indInfoCollection, null indices ]
```

is more easy to read than the new

```haskell
[... | ..., null (resolveEntityIndices False [entity] indInfoCollection) ]
```

but as that's just a matter of taste.

nevrome (approved):
Looks good. Didn't test it myself yet, though.

nevrome:
Ran some test on the command line now and `--ordered` seems to work like a charm :+1:

src/Poseidon/EntityTypes.hs
stschiff: Yes, that was my original plan, but I realised then that `isExcludeEntity` is not a function that one can implement outside the class definition, because it needs to be implemented on a per-instance basis. In a polymorphic function like this one here, it is impossible to inquire about a sign-property without extending the class. I did not want to do this, so I decided for this more general implementation which always "cleans" the existing indices and adds additional ones. I think that's actually more elegant in a way because it is truly agnostic about signed or unsigned entities.

test/PoseidonGoldenTests/GoldenTestsRunCommands.hs
stschiff: Oh good point. Yes, I will add that!

src/Poseidon/EntityTypes.hs
stschiff: That doesn't work anymore, I think. At least it throws a compiler error. I am not sure I fully understand it, to be honest. I agree it looks more elegant, but even before I found this code a bit unintuitive. Usually such a double assignment `[... | a <- as, b <- bs]` is used for double-loops, but here it was used differently. Not sure. In any case, you're welcome to try and turn it back to how it was. I couldn't.

nevrome: I understand. `resolveEntityIndices` previously lived in MonadThrow. Now it's a pure function and this list comprehension now handles it as a normal producer. Ok.

stschiff:
Totally forgot that I hadn't yet merged that. Thanks for the final review.

PR #304, opened 2024-06-28
`--preservePyml`

nevrome:
To conclude the `--ordered` saga I drafted an implementation for the `--preservePyml` argument as discussed in https://github.com/poseidon-framework/poseidon-hs/issues/300. I had to make some design decisions you may or may not like, @stschiff and @TCLamnidis, so please comment.

I went for a simple `--preservePyml` switch that only works in case there is a single relevant source package. If there are more, and the option is active, then `forge` stops with an error.
With `--preservePyml` `forge` runs through normally, but then replaces the following fields of the generated POSEIDON.yml file from the source POSEIDON.yml file:
- `description`
- `contributor`
- `packageVersion`
- `lastModified`
- `readmeFile`
- `changelogFile`

Copying the `readmeFile` and `changelogFile` fields requires of course also to copy the respective files.

You'll note that the package `title` is not copied. It can be easily reproduced by the user with the `-o` or `-n` options.

There is a bit of a weird interaction with the `--minimal` flag. With `--preservePyml` and `--minimal` the resulting package may have a README and a CHANGELOG file, but not a .janno or a .bib file. This sounds like one of these unexpected interactions that could bite us eventually...

I haven't written any tests yet.


stschiff (changes requested):
OK, I checked the code, without trying it out, and I can green light all of these changes. Looks good to me. Would be nice to turn at least one of the existing forge-golden-tests into a test for `--preservePyml`. Would be easy enough I think. 

Regarding the `--minimal` flag, I must admit I never fully understood what use-case that actually supported. I understand why it's in `init`, but for `forge` I never got the point. In any case if we are serious about it, we should not allow `--minimal` and `--preservePyml` together, simple as that: With `--minimal` a user has expectations for empty fields in the Yaml file, while with `--preservePyml` those very fields are filled. These two expectations are fundamentally  incompatible with each other. So I would suggest not to allow that.

nevrome:
I used `--minimal` in the past for pipelines, where I wanted to keep the intermediate data consistently in packages. This was before `--onlyGeno`, though, so maybe it's less relevant now. Still convenient to have it, so I don't want to touch it at the moment.

Anyway: I refactored `forge` to replace the increasingly complex boolean logic by a clear separation of output modes with the new type `ForgeOutMode`. We now have four mutually exclusive output options: `--onlyGeno`, `--minimal`, `--preservePyml` and the default (all covered by golden tests). There are no more unpredictable interactions between them. I think it's much more clear like this.

PR #305, opened 2024-08-02
Gzip reading-support

stschiff:
This PR links to a newer version of sequence-formats, which adds support for reading gzipped genotype files. Here is what I wrote in the Changelog:

- Linked to sequence-formats 1.8.1.0, which adds reading support for gzipped Plink (*.bed and *.bim) and Eigenstrat (.geno and .snp) files.
- gzipped files are recognised automatically by their file ending
- A mild but technically breaking change is the behaviour of init, genoconvert and forge with the `-p, --genoOne` flag, where we now allow only `.bed`, `.bed.gz`, `.geno` or `.geno.gz` files to define the trio of genotype files.


stschiff:
Yes, I think I like this idea. Indeed makes it somewhat clearer for the user, and avoids the breaking change. So just to be clear: You would then say that if the user gives a `*.fam` or a `*.ind` file, we (somewhat arbitrarily) assume that the other two files are unzipped. Correct? Cause that is not covered by the current CLI documentation. The current CLI documentation goes the other way around: If you give a zipped `.geno`, `.snp`, `.bed` or `.bum` file, it assumes the other is also zipped. Anyway, I think I like the solution and will adapt the API.

stschiff:
Oh boy, that was actually a bug! I should have used `takeExtensions` anyway. `takeExtension` can never return `.geno.gz`, so that was actually a bug. Thanks for the unintentional catch!

stschiff:
OK, I have adapted this now as you suggested. So now it's not a breaking change anymore. The code is also quite clear now:

```Haskell
        readGenoInput p = makeGenoInput (dropExtensions p) (takeExtensions p)
        makeGenoInput path ext
            | ext `elem` ["geno",    "snp",   "ind"] = Right (GenotypeFormatEigenstrat, path <.> "geno",    path <.> "snp",    path <.> "ind")
            | ext `elem` ["geno.gz", "snp.gz"      ] = Right (GenotypeFormatEigenstrat, path <.> "geno.gz", path <.> "snp.gz", path <.> "ind")
            | ext `elem` ["bed",     "bim",   "fam"] = Right (GenotypeFormatPlink,      path <.> "bed",     path <.> "bim",    path <.> "fam")
            | ext `elem` ["bed.gz",  "bim.gz"      ] = Right (GenotypeFormatPlink,      path <.> "bed.gz",  path <.> "bim.gz", path <.> "fam")
            | otherwise                              = Left $ "unknown file extension: " ++ ext
```

nevrome:
I added a new test module specifically for CLI interface parsers in 0b6db85, because our golden tests do not cover them (interesting design decision I made back then... but here we are :shrug:).

I think it helped immediately to find and fix a mistake in `-p`. `takeExtensions` grabs the leading `.` before the extensions, and `dropExtensions` removes it. So it must be given explicitly in the extension strings in `readGenoInput`.

stschiff:
Excellent! I also thought that we should test this. Great idea to just start this with one simple interface case.

stschiff:
OK, just to be clear, are you OK with merging this now?

nevrome:
Well - the changelog and the command line documentation must be adjusted after the last changes.

May I ask (again?) why you only implemented reading and not also writing? The feature would have a lot more impact if writing was possible as well.

stschiff:
It's a lot more code, and I prefer to get these features implemented one by one. We don't have to release, of course, but I would like to get this PR in first, and then deal with the rest. I have another PR dealing with VCF-reading which is built on top of this one, which is the next step. After that, I will take care of writing.

stschiff:
I have edited the Changelog. The CLI was already adapted to the new change.

nevrome:
I tested it finally and it worked as expected. So this can be merged, imho. A major version jump is not necessary as this is not breaking any more.

stschiff:
Ah right. Cool, I'll adapt the version and merge it into master then. Thanks!

PR #308, opened 2024-09-06
Vcf support

stschiff:
Draft PR for VCF reading support


stschiff:
OK, this was begun a while ago, and only tests were missing. I've added one now.

Note that as with GZIP, this PR only implements _reading_ of VCF files, not yet of writing. But I think this would already have a number of use cases that are important, so I would vote for getting this into a release even before we support writing. 

There are also a number of decisions I have taken that need to stand the test of time. For example, since VCF does not contain information on group names or Sex, I had to tweak the Janno-Ind consistency check function. Also, VCF reading is somewhat difficult to implement in its full glory of the VCF spec. The Spec is very flexible and complex, and of course my `sequence-formats` library is far from fully implementing a parser for the entire format. It's really somewhat of a minimal parsing that I hope is robust enough to support most use-cases. 

I would like to advertise this feature as "experimental" for now and first have it tested for a while before we move on and review some of the decisions I've made. 

I hope it won't be too much work to review this PR. Happy to give you a run-through in person, @nevrome, if that helps.

src/Poseidon/CLI/Genoconvert.hs
```diff
@@ -98,30 +98,40 @@ convertGenoTo outFormat onlyGeno outPath removeOld inPlinkPopMode outPlinkPopMod
...
+                    GenotypePlink      g _ s _ i _ -> [g, s, i]
+                    GenotypeVCF        g _         -> [g]
+            when removeOld . liftIO . mapM_ (removeFile . (posPacBaseDir pac </>)) $ filesToDelete
   where
     checkFile :: FilePath -> PoseidonIO Bool
     checkFile fn = do
         fe <- liftIO $ doesFileExist fn
         when fe $ logWarning $ "File " ++ fn ++ " exists"
         return fe
+    getFormat :: GenotypeFileSpec -> String
```
nevrome, line 134: This looks like a function that will also be useful elsewhere. Maybe it should be defined where the type is defined.

src/Poseidon/CLI/OptparseApplicativeParsers.hs
```diff
@@ -462,60 +450,59 @@ parseInGenoOne = OP.option (OP.eitherReader readGenoInput) (
...
+                                         <*> parseFileWithEndings "indFile"  [".ind"]              <*> pure Nothing
+    parsePlink      = GenotypeEigenstrat <$> parseFileWithEndings "bedFile"  [".bed", ".bed.gz"]   <*> pure Nothing
+                                         <*> parseFileWithEndings "bimFile"  [".bim",  ".bim.gz"]  <*> pure Nothing
+                                         <*> parseFileWithEndings "bamFile"  [".fam"]              <*> pure Nothing
+    parseVCF        = GenotypeVCF        <$> parseFileWithEndings "vcfFile"  [".vcf", ".vcf.gz"]   <*> pure Nothing
+
+parseFileWithEndings :: String -> [String] -> OP.Parser FilePath
+parseFileWithEndings long endings = OP.option (OP.maybeReader fileEndingReader) (
+    OP.long long <>
+    OP.help ("Accepted file endings are " ++ intercalate ", " endings) <>
```
nevrome, line 491: This "universal" parser function is a good idea. Scales really well. Maybe the "Accepted file endings are ..." help text is a bit too minimalist.

nevrome (commented):
The code looks solid. But there are a lot changes in the CLI subcommands that require good testing. I hope the golden tests cover everything important...

Maybe it's worth to write some more golden tests explicitly for the .vcf input.

I also think the `POSEIDON_*` test data hack is really reaching its limits here. We should have proper test packages for these increasingly different setups!

src/Poseidon/CLI/OptparseApplicativeParsers.hs
stschiff: True. I've now added a bit more specific help texts. Take a look.

stschiff:
Thanks for the review. Golden tests are a good idea, and I will also refactor those test packages.

stschiff:
OK, I've restructured the test packages and added Golden tests. I've also updated the version number, Changelog and Releasechangelog. Have a look.

nevrome:
Perfect! I also ran some little experiments on the command line and could find no issues. So I'll merge this now and trigger the release.

PR #309, opened 2024-09-13
Remove HasColName

stschiff:
OK, this is a quick improvement (in my view) which gets rid of the `HasColName` class, which was neither needed nor secure (things would break if someone defined its instance to actually use `colname`s argument).

Have a look, and if you agree we can quickly merge this in. I will in the mean time review the rest.

Funny story: I just realized today that I think the Aeson-instances in Janno are actually used nowhere across the code-base. I think there might have been an idea to use it for Server communication, but that actually uses a simple type-less encoding for flexibility reasons. Anyway, I suppose it's OK to have them, but right now they are mostly the generic instances, which we have never actually reviewed, or seen in print.


nevrome:
That's a good suggestion - thanks!

PR #310, opened 2024-10-22
removed a lot of unused (?) ToJSON and FromJSON instances

nevrome:
Following the discussion in #307 here's a PR that removes a large number of ToJSON and FromJSON instances. Maybe you could finish the job in `ServerClient.hs`, @stschiff.

If we don't use the instances then I'm all for removing them :slightly_smiling_face:!


stschiff:
From my point this is good to merge into typeMania, where I'll have a final look.

Issue #312, opened 2024-10-29
.janno validation fails for cells with text that ends with an `à` character


nevrome:
@TCLamnidis found an interesting bug in the validation of .janno files. We investigated the issue and eventually understood what is going on.

What Thiseas observed is the following error message when reading a package:

```
Can't read sample in ./***********.janno in line 3577: parse error (Failed reading: conversion error: Cannot decode byte '\xc3': Data.Text.Encoding: Invalid UTF-8 stream in column Location)
```
The offending `Location` field contained a string of the form `******à`.

In our .janno reading procedure we trim whitespaces from `ByteString`s prior to decoding to UTF-8:

https://github.com/poseidon-framework/poseidon-hs/blob/6d6e8f380e86e21655be91b2c614b32f020bb076/src/Poseidon/Janno.hs#L276-L285

This `trimWS` function thus works on raw bytes. It removes every byte at the beginning and end of each .janno cell that fulfills [`isSpace` from Data.Char](https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/src/GHC.Internal.Unicode.html#isSpace), which is defined as follows:

```haskell
isSpace c
  | uc <= 0x377 = uc == 32 || uc - 0x9 <= 4 || uc == 0xa0
  | otherwise = generalCategory c == Space
  where
    uc = fromIntegral (ord c) :: Word
```

Note the `uc == 0xa0`. The character `à` (LATIN SMALL LETTER A WITH GRAVE) is encoded in UTF-8 with the two bytes `\xc3\xa0` (`\195\160`). That means our `trimWS` function removes the second byte of the `à`, if it's at the end of the cell, which breaks the character and causes the UTF-8 decoding in

https://github.com/poseidon-framework/poseidon-hs/blob/6d6e8f380e86e21655be91b2c614b32f020bb076/src/Poseidon/ColumnTypesUtils.hs#L20-L23

to fail. As observed by Thiseas.

My takeaway is that we should not remove whitespaces on the undecoded `ByteString`s, but instead later when the output is properly loaded as `Text`. When making this change we could also test if `removeNoBreakSpace` is still necessary with the recently introduced `Text`-based setup.

Do you think that's sensible, @stschiff?

stschiff:
OK, thanks, seen but not yet dived into. It's on my list.

stschiff:
Great investigation. I just approved #313. Small change big improvement. Thanks.

PR #313, opened 2024-10-30
Fix .janno decoding issues

nevrome:
The bug described in #312 should be fixed. But I also want to see if we can remove `removeNoBreakSpace`. I'll write a test later to check.


nevrome:
If merged this solves #312.

I indeed found we can now remove `removeNoBreakSpace`. I love removing code.

stschiff (approved):
OK, brilliantly simple. And thanks for adding the new tests. Good to merge!

PR #315, opened 2024-11-01
Patch geno one extensions bug

stschiff:
Alright. I think I fixed it, and also added tests.


poseidon-hs.cabal
```diff
@@ -1,5 +1,5 @@
 name:                poseidon-hs
-version:             1.5.7.0
+version:             1.5.7.1
```
nevrome, line 2: I took v1.5.7.1 already for #313. Wer zuerst kommt patched zuerst :grimacing:

poseidon-hs.cabal
```diff
@@ -58,7 +58,8 @@ Test-Suite poseidon-tools-tests
                         Poseidon.SummariseSpec, Poseidon.SurveySpec, Poseidon.GenotypeDataSpec,
                         Poseidon.EntitiesListSpec, PoseidonGoldenTests.GoldenTestsValidateChecksumsSpec,
                         PoseidonGoldenTests.GoldenTestsRunCommands, Poseidon.ChronicleSpec,
-                        Poseidon.SequencingSourceSpec, Poseidon.InterfaceSpec
+                        Poseidon.SequencingSourceSpec, Poseidon.InterfaceSpec,
```
nevrome, line 61: I think `InterfaceSpec` was specifically meant as `OptParseApplicativeParsersSpec`. We only need one module to test code form `OptparseApplicativeParsers.hs`, right? I say: You choose which one you prefer and we remove the other one.

nevrome (approved):
Looks good to me. Sad what we need this complexity, but I don't see how else to deal with with multiple dots in the file names. Thanks for writing the informative tests!

poseidon-hs.cabal
stschiff: OK true. I will take whatever comes next.

poseidon-hs.cabal
stschiff: True, thanks for pointing that out. I will move this over, or merge.

PR #316, opened 2024-11-04
Again allowed 0 in the Nr_SNPs .janno column

nevrome:
This should solve https://github.com/poseidon-framework/poseidon-schema/issues/85

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- V 1.5.7.3:
+    - Allowed `0` in the `Nr_SNPs` .janno column.
 - V 1.5.7.1:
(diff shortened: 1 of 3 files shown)


stschiff:
Hmm, this is OK, but different from what I thought we discussed in https://github.com/poseidon-framework/poseidon-schema/issues/85. I thought we wanted to still issue a warning if the NrSNPs is 0. Now we only get an error if it's negative. That's quite different. Is that on purpose?

nevrome:
Puh - I didn't consider the warning important, honestly. The problem is that the packages with Nr_SNPs < 1 don't get read right now.

Warnings require separate code, because `make` only lives in `MonadFail`. I can add it :man_shrugging:.

Issue #317, opened 2024-11-05
`-p` should allow to select VCF files (both normal and zipped)


nevrome:
Currently the command line documentation for `-p` explicitly excludes VCF files with `For VCF please see option --vcfFile`.

I see no reason for this restriction. `-p` should support reading of `.vcf` and `.vcf.gz` just as `--vcfFile` as a convencience feature.

stschiff:
I have zero emotions here, that's fine. Generally I prefer the "there is exactly one way to do something" philosophy, both for teachability and maintenance, but if that's something that people want. No problem.

stschiff:
This was promptly addressed in #318

PR #318, opened 2024-11-08
support for gzipped-writing

stschiff:
Linking to new version of sequence-formats to gain gzip-writing support for Plink and Eigenstrat


stschiff:
Alright. This is ready for review. I've added support for gzipped writing for PLINK and EIGENSTRAT. `genoconvert` and `forge` get a new CLI option `-z`, which automatically zips the genotype matrix file, and the snp-list file. 

With `genoconvert`, I now first write new files with an ending ( `.gconvert` or `.gconvert.gz` respectively), and rename them after creation. This is necessary, because sometimes only some of the new files are converted, others not, and we cannot risk to override a source file while streaming.

stschiff:
Note that the actual zipping-code is all in sequence-formats. The API is exactly the same as before, but now sequence-formats checks the filename and zips the output if the filename ends with `.gz`. 

For `genoconvert` I also made some code-layout changes and broke very long lines. I now realize this makes it harder to review this PR. Apologies!

stschiff:
I've just also addressed #317 in this PR, and also cleaned the code a bit (there were redundant uses of `<.>`).

stschiff:
*Sigh* the CI fails, despite tests going through just fine on my Mac. Any idea, @nevrome ?

stschiff:
Fixed it. Ready for review

src/Poseidon/CLI/Forge.hs
```diff
@@ -184,15 +185,23 @@ runForge (
                 Nothing -> snpSetMergeList snpSetList intersect_
                 Just _  -> SNPSetOther
     -- compile genotype data structure
-    genotypeFileData <- case outFormat of
-            "EIGENSTRAT" -> return $
+    genotypeFileData <- case (outFormat, outZip) of
```
nevrome, line 188: A bit verbose. Maybe sth like this would work?
```
gz <- if outZip then ".gz" else ""
...
GenotypeEigenstrat (outName <.> "geno" <.> gz)  Nothing
```
This would save seven lines, I think.

src/Poseidon/CLI/Genoconvert.hs
```diff
@@ -42,89 +44,119 @@ data GenoconvertOptions = GenoconvertOptions
...
     -- compile file names paths
     let outName = getPacName . posPacNameAndVersion $ pac
-    (outInd, outSnp, outGeno) <- case outFormat of
-            "EIGENSTRAT" -> return (outName <.> ".ind", outName <.> ".snp", outName <.> ".geno")
-            "PLINK"      -> return (outName <.> ".fam", outName <.> ".bim", outName <.> ".bed")
-            _  -> liftIO . throwIO $ PoseidonGenericException ("Illegal outFormat " ++ outFormat ++ ". Only Outformats EIGENSTRAT or PLINK are allowed at the moment")
-    -- check if genotype data needs conversion
-    if getFormat (genotypeFileSpec (posPacGenotypeData pac)) == outFormat
-    then logWarning "The genotype data is already in the requested format"
+    (outInd, outSnp, outGeno) <- case (outFormat, outZip) of
```
nevrome, line 82: Same trick should work here as well.

src/Poseidon/CLI/Genoconvert.hs
```diff
@@ -42,89 +44,119 @@ data GenoconvertOptions = GenoconvertOptions
...
+                            ". Only Outformats EIGENSTRAT or PLINK are allowed at the moment"
+                runEffect $ eigenstratProd >-> printSNPCopyProgress logA currentTime >-> outConsumer
+            ) (throwIO . PoseidonGenotypeExceptionForward errLength)
+        -- the following will just override if the file already exists.
+        liftIO $ renameFile (outG ++ ".gconvert" ++ zipEnding) outG
+        liftIO $ renameFile (outS ++ ".gconvert" ++ zipEnding) outS
+        liftIO $ renameFile (outI ++ ".gconvert") outI
+        logInfo "Done"
+        -- overwrite genotype data field in POSEIDON.yml file
+        unless (onlyGeno || isJust outPath) $ do
```
nevrome, line 137: We discussed at some point that these options are redundant. And their command line documentation is not very helpful. Maybe we should remove `--onlyGeno` and describe the effect of `-o` a bit better.

src/Poseidon/CLI/Genoconvert.hs
```diff
@@ -42,89 +44,119 @@ data GenoconvertOptions = GenoconvertOptions
...
+                        GenotypePlink      outGeno Nothing outSnp Nothing outInd Nothing
+                    _  -> liftIO . throwIO . PoseidonGenericException $
+                        "Illegal outFormat " ++ outFormat ++
+                        ". Only Outformats EIGENSTRAT or PLINK are allowed at the moment"
+            let oldGenotypeData = GenotypeDataSpec gFileSpec (genotypeSnpSet . posPacGenotypeData $ pac)
+                newPac = pac { posPacGenotypeData = oldGenotypeData }
+            logInfo "Adjusting POSEIDON.yml..."
+            liftIO $ writePoseidonPackage newPac
+        -- delete now replaced input genotype data
+        let oldBaseDir = posPacBaseDir pac
```
nevrome, line 151: This section is a bit odd, because none of the code has to run unless `removeOld`. Maybe it makes sense to refactor this, so that this condition is checked first. I just think it would be easier for us to read.

src/Poseidon/CLI/OptparseApplicativeParsers.hs
```diff
@@ -446,35 +446,37 @@ parseInGenoOne = OP.option (OP.eitherReader readGenoInput) (
...
         readGenoInput :: FilePath -> Either String GenotypeFileSpec
         readGenoInput p =
             let (path, extension) = splitExtensionsOptGz p
             in  makeGenoInput path extension
         makeGenoInput path ext
             | ext `elem` [".geno",    ".snp",   ".ind"] =
-                Right $ GenotypeEigenstrat (path <.> ".geno")    Nothing
-                                           (path <.> ".snp")     Nothing
-                                           (path <.> ".ind")     Nothing
+                Right $ GenotypeEigenstrat (path <> ".geno")    Nothing
```
nevrome, line 463: But... why? `<.>` is neat...

nevrome (commented):
Just looked through the code so far. Didn't play with it yet.

src/Poseidon/CLI/Forge.hs
stschiff: Absolutely! Fixed!

src/Poseidon/CLI/Genoconvert.hs
stschiff: Indeed. Also fixed. Thanks.

src/Poseidon/CLI/Genoconvert.hs
stschiff: Ja, können wir machen. Wäre dann halt streng-genommen ein breaking change

src/Poseidon/CLI/Genoconvert.hs
stschiff: Yes indeed. I've refactored it. Thanks.

src/Poseidon/CLI/OptparseApplicativeParsers.hs
stschiff: True. Don't know why I opted for that. I've brought it back now.

stschiff:
Thank you. I've addressed all the code-comments so far.

src/Poseidon/CLI/Genoconvert.hs
nevrome: Ok - you're right. I bumped to 1.6.0.0 now instead of 1.5.8.0. Version numbers are for free, after all.

nevrome:
OK - I ran some tests on the command line. Works super well, overall! I just found some unintuitive behaviour regarding back-conversions. Here's what I tried:

- Merging two packages with `forge` and creating a zipped output package in PLINK format
- Converting this zipped package to EIGENSTRAT format with `genoconvert`
- Converting it back to zipped PLINK format. 
:warning: Here I observed that the conversion failed (as expected with `Package already in desired file-type`), but surprisingly the POSEIDON.yml file was not adjusted to the .gz files again!
- Converting normal genotype data to the zipped form in both PLINK and EIGENSTRAT format both to a new and the source directory
- Converting it back within the source directory
:warning: Here I got the same error message as for packages (`Package already in desired file-type, skipping genotype conversion for *2010_RasmussenNature-0.1.0*`). That is confusing, because I was not operating on a Poseidon package, but only on unpackaged genotype data (`-p`) here. So while the behaviour is correct, the warning should be different.

So I only saw issues with the back-conversion. Here I think the following behaviour and error messages would be desirable:

If a package is back-converted to an already available format the currently given warnings
```
[Warning] File ./2010_RasmussenNature.bed exists
[Warning] File ./2010_RasmussenNature.bim exists
[Warning] File ./2010_RasmussenNature.fam exists
[Warning] Package already in desired file-type, skipping genotype conversion for *2010_RasmussenNature-1.0.0*
```
are printed, but the POSEIDON.yml file still gets adjusted with a concluding `[Info]` message. Maybe a bit like this:

```
[Warning] File ./2010_RasmussenNature.bed exists
[Warning] File ./2010_RasmussenNature.bim exists
[Warning] File ./2010_RasmussenNature.fam exists
[Warning] Package already in desired file-type, skipping genotype conversion for *2010_RasmussenNature-1.0.0*
[Info] Adjusting POSEIDON.yml file...
[Info] Done
```

If unpackaged genotype data is back-converted to a form that is already there (at least by file name), then genoconvert should not do anything -- as already implemented. But it should also report that with a message (maybe on the `[Error]` level) to indicate that the command did not perform any actions:

```
[Warning] File ./2010_RasmussenNature.bed exists
[Warning] File ./2010_RasmussenNature.bim exists
[Warning] File ./2010_RasmussenNature.fam exists
[Error] No files were created or overwritten
```

What do you think, @stschiff? I did not check how practical it would be to implement this...

stschiff:
Yes, excellent summary of what _should_ be happening, both in terms of actual behaviour and in terms of log-messages. I will look into how to implementing this. Should not be too difficult.

stschiff:
OK, I've implemented the necessary changes for your desired behaviour, @nevrome. I've also run the test scheme that you proposed above. Seems to work as expected. Take a look.

nevrome:
I tested all sorts of back and forth conversions. Worked like a charm!

The only thing I see now is that you decided to make this change a major version upgrade. And I wonder, if it should only be a minor one, because it only adds a new feature without breaking previously available functionality. But maybe I'm missing something.

Issue #321, opened 2024-11-21
trident init issue


guevara2024:
Hi, I am having issues with trident init. It creates an empty output directory and shows an error.

This is what I run:
trident init \
  --genoFile /path/to/file/myFile.geno \
  --snpFile /path/to/file/myFile.snp \
  --indFile /path/to/file/myFile.ind \
  --snpSet 1240K \
  -o /path/to/output \
  -n output

And this is the error I get:
`ERROR: [Error]   File /path/to/file/myFile does not exist`

stschiff:
Thanks for reporting. We have a quick fix in the pipeline!

nevrome:
This will be fixed in the next release.

PR #322, opened 2024-11-21
Fix `parseInGenoSep`

nevrome:
Ui - `parseInGenoSep` has two bugs (I think): It removes needed file extensions (#321) and it reads PLINK data as `GenotypeEigenstrat`.

This is not a testament to good quality control... I really think we made a mistake with the way we set up our golden tests without the actual optparse-applicative interface.

Please check thoroughly to make sure we get it right now!


stschiff (approved):
Oh drat. OK, thanks for fixing. Looks all good! Do you want to merge this maybe into the gzip-writing branch? Then we save one version update? Or merge immediately and update the Version?

Issue #298, opened 2024-05-24
trident rectify could have an option to check and fill the `Nr_SNPs` Janno-column 


stschiff:
see title.

nevrome:
Could be implemented together with #343.

PR #326, opened 2025-01-03
Added a feature to remove empty .janno columns with rectify

nevrome:
I started to think about #298 and realized `rectify` doesn't touch .janno files at all so far. So I thought I first add a simpler .janno-editing feature: Removing empty columns. That is also a pretty common task. Maybe one day `rectify` will rather be `modify` :thinking:


src/Poseidon/Janno.hs
```diff
@@ -394,15 +395,41 @@ createMinimalSample (EigenstratIndEntry id_ sex pop) =
...
 
+-- | A helper functions to replace empty bytestrings values in janno files with explicit "n/a"
+explicitNA :: Bch.ByteString -> Bch.ByteString
+explicitNA = replaceInJannoBytestring Bch.empty "n/a"
+
+replaceInJannoBytestring :: Bch.ByteString -> Bch.ByteString -> Bch.ByteString -> Bch.ByteString
+replaceInJannoBytestring from to tsv =
+    let tsvRows = Bch.lines tsv
+        tsvCells = map (Bch.splitWith (=='\t')) tsvRows
+        tsvCellsUpdated = map (map (\y -> if y == from || y == Bch.append from "\r" then to else y)) tsvCells
```
stschiff, line 406: What's going on with "\r"? Why would that happen?

src/Poseidon/Janno.hs
```diff
@@ -394,15 +395,41 @@ createMinimalSample (EigenstratIndEntry id_ sex pop) =
...
+
+writeJannoFileWithoutEmptyCols :: FilePath -> JannoRows -> IO ()
+writeJannoFileWithoutEmptyCols path (JannoRows rows) = do
+    let jannoAsBytestring = Csv.encodeByNameWith encodingOptions (makeHeaderWithAdditionalColumns rows) rows
+        jannoAsBytestringwithNA = explicitNA jannoAsBytestring
+    case Csv.decodeWith decodingOptions Csv.NoHeader jannoAsBytestringwithNA :: Either String (V.Vector (V.Vector Bch.ByteString)) of
+        Left _  -> error "internal error, please report"
+        Right x -> do
+            let janno = V.toList $ V.map V.toList x
+                jannoTransposed = transpose janno
```
stschiff, line 428: Clever! I didn't know Cassava can simply parse into a vector of vectors. That's great. I think in that case we could just improve `replaceInJannoBytestring` above using that same parsing code, to catch empty strings? Or build it into here.

stschiff (approved):
Very nice. I haven't tested it, but the unit tests look convincing. I left a few comments about the empty-string to "n/a" feature.

src/Poseidon/Janno.hs
```diff
@@ -394,15 +395,41 @@ createMinimalSample (EigenstratIndEntry id_ sex pop) =
...
 -- Janno file writing
 
+-- | A helper functions to replace empty bytestrings values in janno files with explicit "n/a"
+explicitNA :: Bch.ByteString -> Bch.ByteString
+explicitNA = replaceInJannoBytestring Bch.empty "n/a"
+
+replaceInJannoBytestring :: Bch.ByteString -> Bch.ByteString -> Bch.ByteString -> Bch.ByteString
+replaceInJannoBytestring from to tsv =
+    let tsvRows = Bch.lines tsv
+        tsvCells = map (Bch.splitWith (=='\t')) tsvRows
```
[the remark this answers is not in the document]
nevrome: I know - I didn't write this function for this PR, but only moved it within the code. It's old.

I'll think about how to replace it - either with the more clever decoding-encoding mechanism introduced in this PR, or with more targeted encoding somewhere upstream. I'll propose something.

src/Poseidon/Janno.hs
nevrome: No idea - I didn't touch this code beyond moving it around. I'll have a look.

src/Poseidon/Janno.hs
nevrome: Right - that would be one possible solution.

src/Poseidon/Janno.hs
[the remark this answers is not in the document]
nevrome: Ok - I think I found a much cleaner solution to make the NA values explicit.

src/Poseidon/Janno.hs
nevrome: I managed to remove this function entirely.

nevrome:
I found a better solution for `explicitNA`. This operates more upstream, thus also affecting the output of `list`. Please have another look.

stschiff (approved):
OK, this looks all good! Thanks. I think I'd now also like to add a feature in `list` to skip empty columns.

PR #328, opened 2025-01-14
Switch to GHC 9.6.6 and a new resolver

nevrome:


stack.yaml
```diff
@@ -1,14 +1,9 @@
-resolver: lts-21.17
+resolver: lts-22.43
 
 packages:
 - .
 extra-deps:
```
nevrome, line 5: @stschiff I boldly removed some of these fixed versions. I only kept the ones where I remember why we fixed them originally. The tests run through, so I guess this is fine, right?

stschiff (approved):
Excellent, thanks. I've looked at the code changes, and I've just tested it on my Mac with the new GHC, including pedantic tests. Looks all good to me.

Issue #331, opened 2025-01-27
ASCII-only for PoseidonID and GroupNames


stschiff:
I just tried to change both the Janno column types for `PoseidonID` and for GroupNames` to `ByteString.Char8` within the new VCF-writing PR, but got weird compiler errors around `makeInstances`. I don't know template Haskell, so can't continue on this. I will for now leave these types as `String` and `Text`, respectively, and would need some help to get this done.

On the `sequence-formats` side, we now use exclusively ByteString.Char8 for these ASCII-restricted fields.

stschiff:
I would also like to add that I have - for now - decided to _not_ implement checks on the level of the `sequence-formats` library. I looked into it and I noticed that there are many many data formats in that library where I'm parsing into `ByteString`s, so no Unicode support. Setting up checks for all of these is not possible, and feels a bit arbitrary now to implement checks at that level only for individual and group names. It feels then that `sequence-formats` is not the right place for this. 

`sequence-formats` is a low-level library to parse these Bioinformatic formats and constraints for the Encoding are not really a question of low-level parsing, but of client code. So I would like to implement Unicode-checks at the level of `poseidon-hs`, not of `sequence-formats`.

nevrome:
Let's continue this discussion in #332

PR #335, opened 2025-02-04
Vcf writing support

stschiff:
Genoconvert and forge now support VCF as output format. Also zipped works.


stschiff:
Alright. This is ready for review. I have added automatic tests, and have also tested it manually, which already made me change one warning-behavior, which was previously per-SNP. For VCF output, I need to know GenotypePloidy from the Janno File. If that is not set, I _assume_ diploid genotypes, and to make this clear, I output a warning for every individual with missing GenotypePloidy. The rest is all pretty clear, I would hope.

src/Poseidon/CLI/Genoconvert.hs
```diff
@@ -98,10 +103,15 @@ convertGenoTo outFormat onlyGeno outPath removeOld outPlinkPopMode outZip pac =
             liftIO $ createDirectoryIfMissing True (dropTrailingPathSeparator x)
             return x
         Nothing -> return $ posPacBaseDir pac
-    let (outGabs, outSabs, outIabs) = (newBaseDir </> outGrel, newBaseDir </> outSrel, newBaseDir </> outIrel)
+    let outFilesAbs = map (newBaseDir </>) outFilesRel
+    let outFilesAbsTemp = do -- loop over files
+            f <- outFilesAbs
+            if drop (length f - 3) f == ".gz" then
```
nevrome, line 109: That's a bit hacky... Maybe it would be more easily readable with `takeExtension`?

src/Poseidon/GenotypeData.hs
```diff
@@ -355,3 +368,59 @@ printSNPCopyProgress logA startTime = do
...
+
+writeVCF :: (MonadSafe m) => LogA -> [JannoRow] -> FilePath -> Consumer (EigenstratSnpEntry, GenoLine) m ()
+writeVCF logA jannoRows vcfFile = do
+    let sampleNames = map (B.pack . jPoseidonID) jannoRows
+        groupNames  = map ((\(GroupName n) -> T.unpack n) . head . getListColumn . jGroupName) jannoRows
+    forM_ jannoRows $ \jannoRow -> do
+        when (jGenotypePloidy jannoRow == Nothing) . logWithEnv logA . logWarning $
+            "Missing GenotypePloidy for individual ++ " ++ jPoseidonID jannoRow ++
+            ". For VCF output I will assume diploid genotypes. " ++
+            "Please set the GenotypePloidy column explitly in the Janno File to Haploid or Diploid."
```
nevrome, line 380: better:
> Please set the Genotype_Ploidy column explitly in the .janno file to haploid or diploid.

To get the column and value names right: `Genotype_Ploidy`, `haploid`, `diploid`.

src/Poseidon/Janno.hs
```diff
@@ -347,9 +347,9 @@ createMinimalJanno xs = JannoRows $ map createMinimalSample xs
 createMinimalSample :: EigenstratIndEntry -> JannoRow
 createMinimalSample (EigenstratIndEntry id_ sex pop) =
     JannoRow {
-          jPoseidonID                   = id_
+          jPoseidonID                   = Bchs.unpack id_ -- TODO: this will have to change. We need to make PoseidonID itself ByteString
         , jGeneticSex                   = GeneticSex sex
-        , jGroupName                    = ListColumn [GroupName $ T.pack pop]
+        , jGroupName                    = ListColumn [GroupName . T.pack . Bchs.unpack $ pop] -- same thing, see above.
```
nevrome, line 352: I guess that's related to https://github.com/poseidon-framework/poseidon-hs/issues/331

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac21/ForgePac21.vcf
```diff
@@ -0,0 +1,24 @@
+##fileformat=VCFv4.2
+##source=trident_v1.7.0.0
```
nevrome, line 2: Having this line in the golden test is a bit of a headdache. Will need to change with every version change...

nevrome (commented):
Looked through the code. Looks good :+1:

Will try it out next.

nevrome:
One first observation from testing. The command line documentation in `parseOutGenotypeFormat` must now of course include `VCF`.

I find it actually pretty suprising that we only work with strings here: `parseOutGenotypeFormat :: Bool -> OP.Parser String`. Why did we never introduce a type for these formats? :open_mouth:

nevrome:
I found an odd issue during testing. Starting from good old `2010_RasmussenNature` I ran

```
trident genoconvert -d . --outFormat VCF --zip
```

That worked well. I then tried to convert to VCF with

```
trident genoconvert -d . --outFormat VCF
``` 

This failed with

```
[Info]    1 found
[Info]    Checking Poseidon versions...
[Info]    Packages loaded: 1
[Info]    Unpackaged genotype data files loaded: 0
[Info]    Converting genotype data in *2010_RasmussenNature-2.1.1* to format "VCF":
[Info]    Processing SNPs...
[Info]    SNPs:         0    0s
[Info]    Probing of the first 1000 SNPs successful. Continuing now...
[Error]   Issues in genotype data parsing: SeqFormatException "Invalid Alternative Allele"
```

```
trident genoconvert -d . --outFormat EIGENSTRAT
```

unsurprisingly failed with the same error.

In my testing I managed to get out of this locked state twice with 

```
trident genoconvert -d . --outFormat EIGENSTRAT --zip
```

and from there I could then successfully run

```
trident genoconvert -d . --outFormat VCF
```

But I could not consistently reproduce this.

Can you confirm this issue? Any idea what may be the reason?

nevrome:
I ran [vcf-validator](https://vcftools.github.io/perl_module.html#vcf-validator) on a number of trident-generated .vcf files. I saw the following three types of messages:

```
The header tag 'reference' not present. (Not required but highly recommended.)
```
```
The header tag 'contig' not present for CHROM=1. (Not required but highly recommended.)
```
```
1:13740322 .. REF allele listed in the ALT field??
```

stschiff:
Goodness.... good catch! OK, there seems to be something wrong with the VCF. I will investigate.

stschiff:
OK, so the respective site has an `N` as Reference and Alternative allele: 

```
1       12912088        rs12734121      A       G       .       PASS    NS=1    GT      0
1       12913896        rs11810835      T       C       .       PASS    NS=1    GT      1
1       13740322        rs61780219      N       N       .       s10;s50 NS=0    GT      .
1       13762578        rs71629813      N       N       .       s10;s50 NS=0    GT      .
1       13775122        rs11579830      G       A       .       PASS    NS=1    GT      1
```

I will see how that happened.

stschiff:
Yep, the BIM file also has these two Ns. Hmm...

src/Poseidon/CLI/Forge.hs
```diff
@@ -188,15 +189,16 @@ runForge (
     let gz = if outZip then "gz" else ""
     genotypeFileData <- case outFormat of
             "EIGENSTRAT" -> return $
-                GenotypeEigenstrat (outName <.> ".geno" <.> gz)  Nothing
-                                   (outName <.> ".snp" <.> gz)  Nothing
-                                   (outName <.> ".ind") Nothing
+                GenotypeEigenstrat (outName <.> "geno" <.> gz)  Nothing
+                                   (outName <.> "snp" <.> gz)  Nothing
+                                   (outName <.> "ind") Nothing
```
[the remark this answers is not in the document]
stschiff: No, we used to have dots in these extensions, despite the `<.>` already adding a dot. It turns out that `<.>` is smart in the sense that it makes sure that no two dots are added if the extension already contains one. It does feel more "correct" to remove the dot, so that's why I fixed this here.

src/Poseidon/CLI/Genoconvert.hs
stschiff: Yeah, absolutely, no idea what I was thinking. Rewritten with `takeExtension` as suggested. Thanks.

src/Poseidon/GenotypeData.hs
stschiff: Absolutely. Done.

src/Poseidon/Janno.hs
stschiff: Yes.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac21/ForgePac21.vcf
stschiff: Yes, I was seriously considering not printing the version because of that. But I think it's actually nice to have it, in principle. Maybe I need to remove this line then from the outputing VCF file...

stschiff: OK, I've looked into our Golden test and it's actually pretty painful to make any manual changes to the files before they're written out. So I think either we want to keep the feature that trident outputs its own version into the VCF, then I guess we have to live with the fact that the Golden-test VCF files will change with every version, or we remove that feature from the VCF output. I can live with both, honestly. What do you think @nevrome ?

stschiff:
OK, I think I have resolved all comments. Many thanks for the thorough testing, and for the helpful vcf-validator run, which made this a quick fix for me!

Regarding this comment:
> I find it actually pretty suprising that we only work with strings here: parseOutGenotypeFormat :: Bool -> OP.Parser String. Why did we never introduce a type for these formats? 😮

I agree and have now introduced a new datatype `GenotypeOutFormatSpec`, which makes the code safer and actually I could remove some error throws.

The only bit that I have left unresolved is your [comment above](https://github.com/poseidon-framework/poseidon-hs/pull/335#discussion_r1960105058) about the trident version now being output into the Golden Test VCFs. I commented on this above, and I don't know what's the best solution. If you also don't know, I can remove the version output to make life simpler for us.

test/PoseidonGoldenTests/GoldenTestData/forge/ForgePac21/ForgePac21.vcf
nevrome: I think it's good to have the trident version in there. Could be super helpful with debugging. So let's just accept this minor nuisance for now.

Maybe one day we can find a way to avoid this issue in the golden tests. Some chronicle/timetravel-related files also change all the time, so we already have a TODO there anyway.

nevrome:
OK - excellent work! Feel free to merge this.

Just one question: Do we even have to increment the major version here? You just add a feature, right? Or does this PR break anything minor?

Issue #330, opened 2025-01-21
trident rectify --addNewContributor allow ORCID ID


TCLamnidis:
The current DSL for `--addNewContributor` allows for Name(email), but not ORCID ID, even though that is available as a field in the YML file. 

We should update the DSL to allow ORCID IDs (maybe `Name(email)<ORCID>`?)

nevrome:
This was implemented in #396.

PR #337, opened 2025-02-07
Rectify now checks diffs before updating the package version and Changelog

stschiff:
I recently used `trident rectify` to bulk update dozens of packages (for [this PR](https://github.com/poseidon-framework/community-archive/pull/247)) and noticed that when running it on a large base directory, it just updates every single package, regardless of whether anything in the YAML actually changed.

I changed this behaviour now: It now checks whether the YAML actually got updated before increasing the package version and adding to the changelog.

I think this makes more sense.


nevrome:
I see - so this brings back some of the original functionality of this subcommand. Could indeed be useful for bulk updates!

But the way it is implemented now makes it for example impossible to increase the package version, if no other change occured, right? Let's say I change some package file, update the checksum manually, and then want to update version and changelog with `rectify`? This usecase is not covered any more with your `when (inPac /= updatedPacChecksums)` check, correct?

Could we maybe bring this back? We could add an argument like `--onlyChanged` to cover your desired behaviour, but leave the default as it was?

***

The current `rectify` is a bit stupid in the sense that every single change to the package is opt-in. One good example is `--jannoRemoveEmpty`. It does NOT automatically update the .janno checksum, although that may *usually* be necessary. Instead its documentation states:

> Reorder the .janno file and remove empty colums. Remember to pair this option with --checksumJanno to also update the checksum.

I like this stupidity, because it gives you a lot of control. It's orthogonal to the original `trident update` that tried to be a clever wizard.

stschiff:
Hmm, OK, so to your first point: I'm not quite following the very strange hybrid use case you describe. For me there are two reasonable ways to update packages:
1) Manually open an editor, compute checksums, make manual updates, update the version number and changelog. Save everything. Done.
2) Use `trident rectify` to update checksums, logtext and version automatically and programmatically.

Why would someone go through the trouble and starting with the manual process (1) of opening the YAML file with an editor, make the complicated step of manually computing a checksum but then falling short on the comparably trivial update of the version, and instead then rely on (2) to make that trivial step? I think that makes no sense. I see no good reason why `trident rectify` should update a package version if it didn't actually change anything. 

What I understand is that this breaks previous behavior, so perhaps we should at least implement a `--force` option or so that enforces an update despite no change. 

I think your proposal of an `--onlyChanged` option is something I would agree to for pragmatic reasons, but only if I cannot convince you of this any further :-). 

Perhaps it would be good if @TCLamnidis could briefly comment on this, given his extensive use of `trident rectify` inside the Minotaur workflow?

nevrome:
> Why would someone go through the trouble...

I've been there. Changing a checksum is easy and I often do it by hand, but editing the changelog is cumbersome. It involves many more steps: Creating a correctly formatted (!) file, referencing it in the .yml file, incrementing the version number, using the correct version number in the changelog. So I rather let `rectify` do this second step for me.

I prefer a `rectify` that "just does the thing" without any conditions that I have to remember and wonder about later on :shrug:. We've already tried a wizard version and it confused people, me included.

I think `--onlyChanged` is a good compromise. But `--force` would be tolerable as well, I guess.

stschiff:
OK then. I think, though, the `-onlyChanged` option then applies only if the user selected `--packageVersion`. I'll think of a good CLI-API for that then.

stschiff:
OK, I think I've solved it the way discussed. Ready for review.

TCLamnidis:
I share in Stephan's perplexity regarding option 1, but https://imgs.xkcd.com/comics/workflow.png

Between `--force` and `--onlyChanged` I prefer `--force`, as it is the less error-prone approach. Typing the command and pressed Enter prematurely? No biggie, rectify only touches the minimal applicable set of packages. 

The alternative would mean having to roll back a bunch of changes and retrying. Much higher cost of accidentally sending the right command from the wrong directory. 

At least so I'm told... Definitely never happened to me, and I certainly have never spent 15 minutes rolling back accidental updates to all PMA packages because I ran rectify in the wrong directory... 🥹 😢 👀

nevrome:
Now that is an argument that indeed makes sense to me :thinking:. This exact thing has also happend to me as well. And I imagine this must be a bigger pain if no version control allows for easy rollbacks.

You already implemented the other solution, Stephan, but I leave it up to you if you want to change it once more.

stschiff:
I've tested it. Works nicely. Yeah for secure bulk-rectifies!

nevrome:
OK - @stschiff I feel bad for opening this again, but after playing around with this version and looking into the code, I fear we have not reached an optimum yet.

`force` should not only affect version updates. It was just the example raised above. But I think e.g. contributors and potentially other changes should also be enforcable. `--force` should essentially preserve the old behaviour before this PR. I think the state you have reached at c01e301 was fine. Maybe we should go from there, and just add force to the new condition?

Instead of

```haskell
when (inPac /= updatedPacChecksums) $ do
```

we could just have

```haskell
when ((inPac /= updatedPacChecksums) || force) $ do
```

That plus the changes you made to the contributors API and we're good, I think. Maybe that would be an opportunity to quickly address #330 while we're at it.

From this state I would then suggest to investigate #338.

One additional comment: `when ((inPac /= updatedPacChecksums)` has one weakness. I noticed this when I ran the current version on the entire community archive with `trident rectify -d .`, expecting no change. Unfortunately three POSEIDON.yml files where, in fact, overwritten.

<details>
<summary>relevant diff</summary>

```diff
 poseidonVersion: 2.7.1
 title: 2019_Villalba_Iberia
-description: Survival of Late Pleistocene Hunter-Gatherer Ancestry in the Iberian Peninsula.
-  Originally AADR v42.4. Peninsula.
+description: Survival of Late Pleistocene Hunter-Gatherer Ancestry in the Iberian
+  Peninsula. Originally AADR v42.4. Peninsula.
 contributor:
 - name: Vanessa Villalba
   email: villalba@shh.mpg.de
diff --git a/2020_Nakatsuka_Andes/POSEIDON.yml b/2020_Nakatsuka_Andes/POSEIDON.yml
index 8ae2f66f..b0894108 100644
--- a/2020_Nakatsuka_Andes/POSEIDON.yml
+++ b/2020_Nakatsuka_Andes/POSEIDON.yml
@@ -1,6 +1,7 @@
 poseidonVersion: 2.7.1
 title: 2020_Nakatsuka_Andes
-description: Transect of ancient samples from the Andes covering 10000 years. Originally AADR v42.4.
+description: Transect of ancient samples from the Andes covering 10000 years. Originally
+  AADR v42.4.
 contributor:
 - name: Ayshin Ghalichi
   email: ghalichi@shh.mpg.de
diff --git a/2023_Penske_EarlyContact/POSEIDON.yml b/2023_Penske_EarlyContact/POSEIDON.yml
index 6b4107a6..ee47d710 100644
--- a/2023_Penske_EarlyContact/POSEIDON.yml
+++ b/2023_Penske_EarlyContact/POSEIDON.yml
@@ -1,6 +1,7 @@
 poseidonVersion: 2.7.1
 title: 2023_Penske_EarlyContact
-description: Early contact between late farming and pastoralist societies in southeastern Europe
+description: Early contact between late farming and pastoralist societies in southeastern
+  Europe
 contributor:
 - name: Ayshin Ghalichi
   email: ayshin_ghalichi@eva.mpg.de
```

</details>

`updatedPacChecksums` probably includes a reformatted version of the yaml file with the linebreak. But why? Is there a way to write this condition that ignores these changes to long lines?

stschiff:
Just as an update, as we talked on Zoom: This PR does _not_ change the behaviour of rectify in any way with respect to updates other than the package Version and the logText. All other changes, like checkSums, Janno changes or contributors, will always happen in all packages, whether or not `--force` is set.

We have brainstormed whether to change this, discussion still ongoing.

stschiff:
I think we have been overthinking this in our discussion. How about the following behaviour, which (almost) non-breaking:

## If a single package was selected
- "Single mode"
- the old behavior as it was. No check. All changes exactly as the user wants them. 
- Want a checksum update without a version update? `--checksumBla`.
- Want to add a contributor without anything else? `--contributor`
- ... simply the same as it was for everything!

## If multiple packages were selected
- We switch to "Bulk mode".
- This means that contributors, package version and logtext are added only if any of the checksums were changed.
- This behavior can be overwritten by `--forceAll`, in which case Bulk mode = single mode, applied to all packages.

Two minor remaks:
- In bulk mode, without a `--checksum` option, package versions, logtexts and contributors will not change.
- `--forceAll` in single mode has no effect. In some sense, single mode has `--forceAll` implied.

What do you think?

nevrome:
I'm still wondering if a split between a limited `rectify` for one and a straightforward `modify` for the other usecase wouldn't be cleaner. Making a subset of changes conditional on multiple requirements (checksums must have changed and `--checksumX` must be active) just doesn't sound intuitive for me.

But we have reached a stalemate and maybe we just have to try something now. I'm OK with it. Maybe we call `--forceAll` just `--force`.

stschiff:
There is another maybe even more transparent way to implement this:
As soon as multiple packages are selected, the program refuses to do anything and outputs a warning, and asks the user to decide between different "bulk" modes:
`--bulk all` means just run the normal modification on every single package without any conditioning
`--bulk onlyChanged` means what I suggested above, so change only those packages whose checksums were updated

That way it's kind of even more fool-proof and we are a bit closer to the interactive "are you sure" kind of mode that @TCLamnidis proposed?

nevrome:
Stephan and I had another discussion about this. Here's a quick summary of my understanding of the results:

- The old, non-smart `rectify` was useful and should be preserved as is. It should get a new name (e.g. `modify`/`alter`/`manipulate`). It's flaw of editing every package below the base directory should be documented clearly. We could consider extra measures like a warning when editing multiple packages, or even a flag that has to be set in case multiple packages will be affected.
- A new, smart subcommand `rectify` should be added. It automatically corrects checksums and bumps the version of packages whose checksums have changed. It has only four options: `-d` (naturally), `--packageVersion VPART` and `--logText STRING` (mandatory), and `--newContributors DSL` (optional).

stschiff:
Yes, I will take a look and probably kill this PR entirely and start from scratch. I will then copy your message from above into the new PR.

Issue #333, opened 2025-02-04
New poseidon server suggestions


stschiff:
Once we've added the `html` branch with the new archive explorer, here are some suggestions for improvement:

- It would be great for the front pages to list a short sections with "newest packages" or "most recent package updates" or so.
- Alternatively, or in addition to that, it would be nice to show the "last-modified" date for each package in the table
- Lastly, it would be great to have the table dynamic, and sortable.

This is just a wishlist, feel free to ignore for now.

nevrome:
- [ ] The front page should maybe also feature some dashboard-like visualizations (?) about the state of each archive

nevrome:
- [ ] The front page should include a powerful full text search. I think the best way to implement this would be a custom search index system in trident (maybe like [this](https://entropicthoughts.com/search-index-150-lines-haskell)). This could then also support a subcommand `trident search` and other derived applications.

PR #341, opened 2025-04-05
some small improvements to the data explorer webpage

nevrome:
Motivated by a comment of @KathrinNaegele I did some work on the html API this weekend.

As suggested by @stschiff in #333, I ...
-  added per-archive lists of last-modified packages to the main page
-  added the last-modified dates for each package in the package table on the archive page
- made the package and individual lists dynamic and sortable with a new dependency on simple-datatables (essentially DataTables without jQuery)

I haven't tested this yet with the Poseidon server production setup.


stschiff (approved):
Great, I've quickly tested it locally. Seems to all work as expected. Thanks.

stschiff:
Note that CI fails. Note sure what exactly though.

PR #344, opened 2025-04-23
a feature to exclude packages from mapping in the html API

nevrome:
This little change allows to give a list of packages that should not be mapped in the archive view of the `serve` html API. It's admittedly a pretty specific feature, but useful for the aadr-archive:

There we currently have the following perspective:

image

With the changes in this PR we can exclude old AADR versions (just from the overview map):

image

I think removing this massive duplication makes the map more readable and useful. It also cuts down the page's loading time.

The config file for this looks like this:

```yml
archives:
- name: aadr-archive
  paths: [.]
  description: Poseidon AADR Archive (PAA) with a structurally unified version of the AADR dataset repackaged in the Poseidon package format.
  URL: https://github.com/poseidon-framework/aadr-archive
  dataURL: https://github.com/poseidon-framework/aadr-archive/tree/main
  excludeFromMap:
  - AADR_v54_1_p1_1240K_BeyondAncient
  - AADR_v54_1_p1_1240K_EuropeAncient
  - AADR_v54_1_p1_1240K_Modern
  - AADR_v54_1_p1_HO_Modern_not_in_1240K
```


src/Poseidon/CLI/Serve.hs
```diff
@@ -85,11 +86,12 @@ parseArchiveConfigFile :: (MonadIO m) => FilePath -> m ArchiveConfig
...
-    , _archSpecPaths       :: [FilePath]
-    , _archSpecDescription :: Maybe String
-    , _archSpecURL         :: Maybe String
-    , _archSpecDataURL     :: Maybe String
+    { _archSpecName               :: ArchiveName
+    , _archSpecPaths              :: [FilePath]
+    , _archSpecDescription        :: Maybe String
+    , _archSpecURL                :: Maybe String
+    , _archSpecDataURL            :: Maybe String
+    , _archSpecExcludePacsFromMap :: Maybe [String]
```
nevrome, line 94: I'm aware that `Maybe [String]` is ugly, but I think it's necessary to play nicely with aeson (see [here](https://stackoverflow.com/questions/21345365/how-to-omit-empty-lists-using-aeson-derivejson)).

stschiff:
Very useful CLI addition! I'll review ASAP.

src/Poseidon/CLI/Serve.hs
stschiff: I've encountered this before and solved it with a simple one-liner (like [here](https://github.com/poseidon-framework/poseidon-analysis-hs/blob/0e9e2e079b509254f5f3cb1cde08515634d12116/src/Poseidon/Analysis/FStatsConfig.hs#L81)

stschiff (approved):
OK nice. I would definitely always prefer `[a]` over `Maybe [a]` and made a suggestion how to make that work, but it's up to you, I don't mind. The rest looks all good. I haven't tested it, though, but from your PR comment it looks like it works as expected!

nevrome:
OK - that's a good fix. I implemented it.

Issue #350, opened 2025-06-17
Validation of accessions from the GSA fails


TCLamnidis:
Currently, creating a package with data accessions from the [GSA Human](https://ngdc.cncb.ac.cn/gsa-human/) fails at validation with the following error:

```
[Error]   Can't read sample in package/2025_Liu_ShanDong.ssf in line 2: parse error (Failed reading: conversion error: Accession "HRI1054931" not a correct biosample/sample accession)
```

I believe the relevant code is here:
https://github.com/poseidon-framework/poseidon-hs/blob/9aafbec6b5580ba8a3375cb7f6805715c9840c88/src/Poseidon/SequencingSource.hs#L62-L73

The issue seems to be that trident validates the accession numbers based on the criteria of the INDSC. The GSA is not (yet?) part of the INDSC, and as such their accession numbers are not detailed in the [INDSC accession guide](https://ena-docs.readthedocs.io/en/latest/submit/general-guide/accessions.html).

## Solution 1: Update validation parameters to allow GSA Human accessions

GSA documentation in various places (e.g. [here](https://download.cncb.ac.cn/bigd/standards/GSA%20Standards.pdf), or [here](https://ngdc.cncb.ac.cn/gsa/support/standardsGsa) suggests the prefixes of their accession numbers (see also figure below), but these seem to be prefixes for the **GSA** and not **GSA Human**.

![Image](https://github.com/user-attachments/assets/e8918ae7-3741-4aa2-b3ab-a0374bd7542f)

I have not been able to find documentation on the accession number formatting for GSA Human (at least not in English).

Instead I attempted to reverse-engineer the patterns based on a real entry.
Looking at [project entry HRA008755](https://ngdc.cncb.ac.cn/gsa-human/browse/HRA008755), the accession numbers for GSA Human seem to be more like:

| Column 1      | Column 2      | Notes |
| ------------- | ------------- | ----- |
| Project | HRA[0-9]{7} | This seems to be linked to a BioProject accession with prefix PRJ[C][A-Z][0-9]{6} |
| Individual | HRI[0-9]{7} | |
| Sample | HRS[0-9]{7} | This seems to be linked to a BioSample accession with prefix SAM[C][0-9]{7}|
| Experiment | HRX[0-9]{7} | |
| Run | HRR[0-9]{7} | |

## Solution 2: Relax accession number validation

I think it is a good idea to offer some validation of these values, but maybe validation failures should be a warning instead of an error.
[...]

PR #354, opened 2025-07-24
Added a mechanism to report non-breaking issues in individual cells of .janno and .ssf files

nevrome:
Multiple times over the past years we have discussed the requirement to report not just parsing failure when reading .janno or .ssf, but also minor non-breaking issues in individual cells. With the upcoming switch to Poseidon v3.0.0 we will probably encounter more cases where we would like to do this.

From a technical point of view this is not trivial. The cassava library we use for parsing does not operate in IO (rightfully so, I guess) and does not allow logging in the reading process. We already have a slightly awkward mechanism check for cross-column consistency after the reading (see e.g. at the bottom of the `Janno` module) and potentially this could have been extended to individual columns. But I find this inelegant and hard to maintain.

Here I propose an alternative solution with a custom type class `Suspicious` for types that should be inspected for anomalies after parsing.

```haskell
class Suspicious a where
    inspect :: a -> Maybe [String]
```

All .janno and .ssf types are suspicious, so they must all be implement as instances of this class. I did so in this PR, and already came up with some useful instances, e.g. to solve #350:

```haskell
instance Suspicious SSFAccessionIDSample where
    inspect (SSFAccessionIDSample x) =
        case x of
            (INSDCBioSample _) -> Nothing
            (INSDCSample _)    -> Nothing
            i                  -> Just ["sample_accession " ++ show i ++ " is not a correct INSDC \
                                         \biosample/sample accession."]
```

or to demarcate "soft" limits:

```haskell
instance Suspicious JannoDateC14UncalBP where
    inspect (JannoDateC14UncalBP x)
        | x >= 60000 = Just ["Date_C14_Uncal_BP is " ++ show x ++", so >60k years and thus beyond the \
                             \pratical limit of radiocarbon dating."]
        | otherwise = Nothing
```

I'm sure we will come up with more.

All of these inspections are performed in `readJannoFileRow` and `readSeqSourceFileRow` respectively, so directly after parsing each row.

```haskell
-- cell-wise checks
let inspectRes = concat $ catMaybes $ inspectEachField jannoRow
OP.unless (null inspectRes) $ do
    logWarning $ "Value anomaly in " ++ jannoPath ++ " in line " ++ renderLocation ++ ": "
    mapM_ logWarning inspectRes
```

The key function `inspectEachField` here is horrible generics-sop magic looping through all record fields of `JannoRow` and `SeqSourceRow`:

```haskell
inspectEachField :: (Generics.SOP.Generic a, Code a ~ '[ xs ], All Suspicious xs) => a -> [Maybe [String]]
inspectEachField =
      hcollapse
    . hcmap (Proxy :: Proxy Suspicious) (\(I x) -> K $ inspect x)
    . unZ . unSOP . Generics.SOP.from
```

Issues are reported on the command line like this:

```
[Warning] Value anomaly in ./Ning_China.janno in line 21 (Poseidon_ID: JXNTM23):
[Warning] Date_C14_Uncal_BP is 3127000, so >60k years and thus beyond the pratical limit of radiocarbon dating.
[Warning] Nr_SNPs is set to 0, indicating no recorded SNPs in the genotype data.
[Warning] Value anomaly in ./ENAtable.ssf in line 5:
[Warning] sample_accession SAMA6490835 is not a correct INSDC biosample/sample accession.
[Warning] Value anomaly in ./ENAtable.ssf in line 9:
[Warning] study_accession 3 is not a correct INSDC project/study accession.
[Warning] run_accession s is not a correct INSDC run accession.
[Warning] Value anomaly in ./ENAtable.ssf in line 39:
[Warning] read_count is set to -1, which indicates a missing value.
[Info]    Packages loaded: 1
```

I think this is neat. What do you think, @stschiff? I'm also happy to hear suggestions for additional warnings, @TCLamnidis.

### DIFF
--- a/src/Poseidon/CLI/Survey.hs
+++ b/src/Poseidon/CLI/Survey.hs
@@ -1,8 +1,8 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- the following ones are necessary for the generics-sop magic
 {-# LANGUAGE DataKinds         #-}
 {-# LANGUAGE FlexibleContexts  #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GADTs             #-}
-{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeOperators     #-}
 
 module Poseidon.CLI.Survey where
--- a/src/Poseidon/ColumnTypesJanno.hs
+++ b/src/Poseidon/ColumnTypesJanno.hs
@@ -198,7 +198,11 @@ instance Makeable JannoDateC14UncalBP where
             Left e -> fail $ "Date_C14_Uncal_BP can not be converted to Int because " ++ e
             Right (num, "") -> pure $ JannoDateC14UncalBP num
             Right (_, rest) -> fail $ "Date_C14_Uncal_BP can not be converted to Int, because of a trailing " ++ show rest
-instance Suspicious JannoDateC14UncalBP where inspect _ = Nothing
+instance Suspicious JannoDateC14UncalBP where
+    inspect (JannoDateC14UncalBP x)
+        | x >= 60000 = Just ["Date_C14_Uncal_BP is " ++ show x ++", so >60k years and thus beyond the \
+                             \pratical limit of radiocarbon dating."]
+        | otherwise = Nothing
 instance Show JannoDateC14UncalBP where          show (JannoDateC14UncalBP x) = show x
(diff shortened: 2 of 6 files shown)


stschiff:
OK, consider me convinced.

stschiff (approved):
OK, I've gone through all that and I approve. I did not find anything I would have done differently at this point. I think it's a very useful machinery and makes this future-proof for upcoming schema-changes.

Issue #325, opened 2024-12-06
Retiree-Packages


stschiff:
As discussed on December 5th, we may need an option to "retire" packages on the server. This is less trivial than it sounds, since our `trident timetravel` mechanism will automatically (and nightly) fetch all packages, even old ones from our git repository. Even if we remove a package from the repository's master branch, it will remain in the record and therefore it will be fetched by `timetravel`. This is intentional: For backwards-compatibility, we want to continue serving old packages, even if they have been superseded by other packages or terminated. An example is package `2019_BraceNatureEcologyEvolution`, which we removed as it turned out to duplicate another package named `2019_Brace_Britain`. It will however forever live in our record and hence it will be served by our server APIs. It will even get marked as `latest` by clients, because some version of it will be the latest, by definition.

There are a number of ideas that we discussed:

1.) We could simply remove that package from the community-archive chronicle file. This will practically erase all traces of this package, and our nightly `trident timetravel` on the server will stop serving this package at all. This option is easy to implement, but it makes a somewhat opaque manual change that breaks our principle of chronicling all old packages. There is no obvious way to document that change so that we can later understand that one entry is missing. Also: It will break any previous workflows from users who have used that package and will be surprised to not be able to fetch it anymore. I think this option is therefore out! Breaking changes are not acceptable, and for whatever reason a user might have to still use a retired package, we shouldn't prevent them from it.

2.) (this was the option we discussed as final solution) We could define a new "retired_packages" file, which we track on the GitHub-repository of the community-archive, which lists packages that we want to retire. We defined in our discussion what "retired" really means: A retired package is a package that will never be marked as `latest`, not even its formally latest version. One thing to think through here is the fact that currently none of our APIs supports requesting only "latest" packages. Rather: the `/packages` API simply lists all packages, in all versions, and the `/zip_file` API serves any single package, either specified with version, or without, in which case it serves the latest. The whole "latest" classification happens entirely inside of `trident list` and related commands on the _client side_. So in order to get this option implemented, we need to implement a new API option `onlyLatest` for `/packages`, which would then skip a package whose package name is in the "retired_packages" file. This would then also require changing `trident list --remote --packages --onlyLatest` internally, to use this API option instead of determining "latest" packages on the client side. Similar changes might be necessary in `trident fetch --downloadAll`. We might also want to change the behavior of the `/zip_file` API in case a package name is requested without version. In that case we could return an error if a retired package is requested, and only serve it when requested with a specific version.

[...]

PR #356, opened 2025-08-20
Server retired packages

stschiff:
Addresses #325. Here is the Changelog entry:

- V 1.6.8.0:
    - Introduced new "retiredPackage" option in the CLI parsers to specify retired packages for serving. Retired packages are by default ignored in the /packages, /groups, /bibliography and /individuals endpoints, as well as ignored in the archive HTML page of the explorer. However, /zip_file still serves retired packages, so that they can be downloaded, and the retired packages are still available in the per-package Explorer HTML page.


stschiff:
OK I've now tested it and tweaked some more things, for example the explorer page now shows also the correct number of packages if retired ones are excluded. You can always add `?includeRetired` to any API, both JSON and HTML, to include retired. I have tested it with the testPackages, and confirmed that indeed excluding for example `Lamnidis_2018` will exclude both versions of these packages, while `Lamnidis_2018-1.0.0` will exclude only this specific one.

nevrome:
Thanks for addressing this, finally! This is a very important feature for the entire ecosystem!

Before I try it out and look at the code I have a first big question: Shouldn't we turn the CLI options `retiredFile` and `retiredPackages` to a feature of the archive configuration .yml file? So to a single new field `retiredFile` in `ArchiveSpec`?

https://github.com/poseidon-framework/poseidon-hs/blob/987b8d7b8c01d04b6a042eab4a1b1b608fe0bf63/src/Poseidon/CLI/Serve.hs#L88-L95

Filtering out retired packages is something that should imho happen on the archive level, not above.

Ultimately my idea is as follows: Every archive should have a separate list of retired packages, tracked with Git in the respective archive repository (then it's perfectly transparent what was retired when and by whom). If `retiredFile` is a field of `ArchiveSpec`, then we can simply add the respective retired-file paths to the [archive config file](https://github.com/poseidon-framework/internal-docs/blob/main/server_archive_config.yml).

stschiff:
Excellent idea!!! I will get on with that suggestion!

stschiff:
OK, I've made this change now. I've reformatted the `Serve.hs` module a bit, as I was getting confused by its complexity. Not sure it helps that much. I added a Golden Test for the retired feature at the ArchiveConfig level. Looks like it works.

src/Poseidon/CLI/Serve.hs
```diff
@@ -102,30 +101,87 @@ instance FromJSON ArchiveSpec where
         <*> v .:? "URL"
         <*> v .:? "dataURL"
         <*> ((v .:? "excludeFromMap") >>= maybe (return []) return)
+        <*> parseCLIretiredPackages v
```
nevrome, line 104: Noticed when testing: With this setup the `retiredPackages` field is mandatory in the .yml file, just like `name` and `paths`. I think it would be better to have it optional like the other fields.

stschiff: Great catch, indeed! Just fixed it.

stschiff:
OK, so I've fixed all the other small points you've raised (all much appreciated!). Your last point I also agree with. But for simplicity sake I would then suggest to _only_ also a retiredFile option in the YAML, and nothing both a direct listing and a file. So I will just replace the current option "retiredPackages" by a new option "retiredPackagesFile" which would then be a YAML list. YAML has the advantage that we can add comments, over straight text files.

PR #363, opened 2026-01-06
Yaml changes v3

stschiff:
I've added the license file fields and the additional reference assembly fields for genotype data to the Yaml file structure.

### DIFF
--- a/src/Poseidon/Package.hs
+++ b/src/Poseidon/Package.hs
@@ -204,7 +204,7 @@ instance ToJSON PoseidonYamlStruct where
 instance FromJSON LicenseSpec where
     parseJSON = withObject "LicenseSpec" $ \v -> LicenseSpec
         <$> v .:  "name"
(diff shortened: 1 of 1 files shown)


src/Poseidon/Package.hs
```diff
@@ -190,6 +201,19 @@ instance ToJSON PoseidonYamlStruct where
         "changelogFile"   .= _posYamlChangelogFile x
         ]
 
+instance FromJSON LicenseSpec where
+    parseJSON = withObject "LicenseSpec" $ \v -> LicenseSpec
+        <$> v .:  "name"
+        <*> v .:  "url"
```
nevrome: Above you define `licenseURL :: Maybe String`, but here you use `v .:  "url"`. I'm surprised this even compiles. Should be `.:?`, right?

nevrome (approved):
Looks good to me :+1:

PR #366, opened 2026-02-09
added smarter lookahead for pacversion parsing

stschiff:
as discussed in person.


stschiff:
Need to add changelog and version bump

PR #369, opened 2026-03-17
Add xerxes codebase

stschiff:
Draft Pull Request to add Xerxes codebase. Work in progress. To be merged into master after #351 has been fully merged.


stschiff:
You're right, that structure _is_ weird. How about a restructuring:

```
src/Poseidon/Core -> what previously we had in `src/Poseidon`
src/Poseidon/Analysis -> what previously we had in `src/Poseidon/Analysis` and in `src/Poseidon/Generator` (possibly involving some further sub-structure for Generator)
src/Poseidon/CLI/Trident -> what previously was in `src/Poseidon/CLI`
src/Poseidon/CLI/Xerxes -> what previously was in ´src/Poseidon/Analysis/CLI` and ´src/Poseidon/Generator/CLI`
```

so a bit of a radical restructuring, but I think more consistent than what we have now.

nevrome:
Yes - I think that would be neat! Let's implement this, but only after the cool #375 is merged. Otherwise we'll run into the merge conflict of the century :grimacing:

nevrome:
As discussed today #375 will still take a bit longer. So we decided to move forward here. I'll do so asap.

PR #375, opened 2026-03-31
Adding Strandcheck and Strandflip-correction

stschiff:
This PR adds and option to strand-check genotype datasets in all commands that consume joint genotype data.


stschiff:
OK, I open this for code review new. It's probably hard to review, sorry @nevrome. The important points:

- I changed the machinery behind merging SNP sets with an emphasis on pure functions, with  no logging. Any logging happens as a `Left`, which is then caught be the `joinEntryPipe` to report as error and resulting in skipping the SNP. Any silent decisions are not anymore logged. I found this far easier to reason about and to maintain, and it was really unnecessary to bother the user with logging of trivial cases like merging genetic positions or snpIds.
- I changed the reasoning behind `getConsensusSnpEntry`, which now uses two different algorithms to determine the consensus SNPs for the two cases with or without strand flips. This is non-breaking for the default case of no strand flips.
- I simplified the code around `recodeAlleles`, which has always been rather complicated because of inherent complexity of dealing with missing alleles. I think the code is now clearer and easier to maintain, and it should be non-breaking for the case of no strand-flips (unit and golden tests confirm).
- the new option `--strandCheck` is _only_ available in `forge`. I do not allow strand-flips in any other sub-command, as I think that is harder to explain then. If a user suspects strand flips (they will notice because the merged number of SNPs will be very low in case there are strand flips but trident assumes none), they should first forge (and thereby correct strand flips), and then run any additional command. At least I as a user would find it a little bit "too smart", if trident where able to correct strand flips somewhat under the hood implicitly every time it reads in multiple datasets.... but that's something that I'm happy to discuss... would be very easy to add the option also to other sub-commands that use `getJointGenotypeData`.

Finally, I have not battle-tested this new feature yet, and I am planning to reach out to users who have reported strand flips in the past. So while this is ready for review, I would like to delay merging until I've had a chance to see it in practice.

src/Poseidon/CLI/Trident/OptparseApplicativeParsers.hs
```diff
@@ -366,6 +366,12 @@ parseIntersect = OP.switch (
         \defined as missing in those packages which do not have a SNP that is present in another package. \
         \With this option set, the forged dataset will typically have fewer SNPs, but less missingness.")
 
+parseStrandCheck :: OP.Parser Bool
+parseStrandCheck = OP.switch (
+    OP.long "strandCheck" <>
+    OP.help "Whether to allow strand flips in the genotype data. Note that this will remove \
+    \any A/T and G/C SNPs from the data, as for those we cannot determine the correct strand orientation.")
```
nevrome, line 373: About the interface in general, as a slightly oblivious user myself:
Does that mean the default behaviour of trident does not change, if this option is not active? Is this a good default? Should `--strandCheck` not be active by default?

You write
> If a user suspects strand flips (they will notice because the merged number of SNPs will be very low in case there are strand flips but trident assumes none) ...

but I wonder if I myself would actually notice this. Should there not be a warning in these cases? Maybe a little report at the end of the forging operation, that reports the number of lost SNPs and makes suggestions like "use --strandCheck" under certain conditions?

src/Poseidon/CLI/Xerxes/FStats.hs
```diff
@@ -139,7 +139,7 @@ runFstats opts = do
     errLength <- envErrorLength
     blocks <- liftIO $ catch (
         runSafeT $ do
-            eigenstratProd <- getJointGenotypeData logA False relevantPackages Nothing
+            eigenstratProd <- getJointGenotypeData logA False False relevantPackages Nothing
```
nevrome: I understand now what you mean by
> they should first forge (and thereby correct strand flips), and then run any additional command. At least I as a user would find it a little bit "too smart", if trident where able to correct strand flips somewhat under the hood implicitly every time it reads in multiple datasets...

I was wondering which subcommand you had in mind, completely forgetting about `xerxes` 🤦. Good that we have this code in here now, and can not delay this decision!

I'm not sure about it, honestly. But I feel xerxes should at least tell me that there is a potential issue if I run it on malformed data. And if I could run `--strandCheck` here, instead of tediously forging everything beforehand, I would probably appreciate the convenience. 🤔

nevrome: I understand now what you mean by
> they should first forge (and thereby correct strand flips), and then run any additional command. At least I as a user would find it a little bit "too smart", if trident where able to correct strand flips somewhat under the hood implicitly every time it reads in multiple datasets...

I was wondering which subcommand you had in mind, completely forgetting about `xerxes` 🤦. Good that we have this code in here now, and can not delay this decision!

I'm not sure about it, honestly. But I feel xerxes should at least tell me that there is a potential issue if I run it on malformed data. And if I could run `--strandCheck` here, instead of tediously forging everything beforehand, I would probably appreciate the convenience. 🤔

src/Poseidon/Core/GenotypeData.hs
```diff
@@ -299,97 +298,130 @@ vcf2eigenstratPipe = for cat $ \vcfEntry -> do
...
     id_ <- case uniqueIds of
         [i] -> return i
         _ -> do -- multiple Ids: Picking the first rs-number if possible, otherwise the first one.
             let rsIds = filter (isPrefixOf "rs") uniqueIds
-                selectedId = case rsIds of
-                    (i:_) -> i
-                    _     -> head uniqueIds
-            logWithEnv logA . logDebug $
-                "Found inconsistent SNP IDs: " ++ show uniqueIds ++ ". Choosing " ++ show selectedId
-            return selectedId
```
nevrome, line 333: Yes - I think it's OK that these very verbose log messages are gone. Even if a user would encounter an issue here, they would probably not discover it in the haystack of messages.

src/Poseidon/Core/GenotypeData.hs
```diff
@@ -299,97 +298,130 @@ vcf2eigenstratPipe = for cat $ \vcfEntry -> do
...
+-- one genotype source. In addition we cannot tolerate strand-ambiguous pairs (C/G and A/T).
+getConsensusAllelesStrandCheck :: [EigenstratSnpEntry] -> Either String (Char, Char)
+getConsensusAllelesStrandCheck snpEntries = do
+    let allAllelePairs    = [(r, a) | EigenstratSnpEntry _ _ _ _ r a <- snpEntries]
+    case filter (\(a, b) -> not (isMissing a) && not (isMissing b)) $ allAllelePairs of
+        [] -> Left $
+            "When checking for strand flips, I require at least one non-missing allele \
+            \pair to determine the consensus alleles. However, all allele pairs are missing for SNP " ++
+            show (snpId . head $ snpEntries) ++ " at position " ++
+            show (snpChrom . head $ snpEntries) ++ ":" ++ show (snpPos . head $ snpEntries)
```
nevrome:
```haskell
let firstEntry = head $ snpEntries
in ... snpId firstEntry ... snpChrom firstEntry ...
```
Just a stylistic suggestion.

But this also reminds me of something: Newer GHC versions report partial functions like `head` as a warning. If we want to switch to a newer compiler eventually and keep running on the `--pedantic` level, then we will have to get rid of them.

src/Poseidon/Core/GenotypeData.hs
```diff
@@ -299,97 +298,130 @@ vcf2eigenstratPipe = for cat $ \vcfEntry -> do
...
+
+-- this is a different algorithm: If strand flips are allowed, we need to have at least one non-missing allele-pair within
+-- one genotype source. In addition we cannot tolerate strand-ambiguous pairs (C/G and A/T).
+getConsensusAllelesStrandCheck :: [EigenstratSnpEntry] -> Either String (Char, Char)
+getConsensusAllelesStrandCheck snpEntries = do
+    let allAllelePairs    = [(r, a) | EigenstratSnpEntry _ _ _ _ r a <- snpEntries]
+    case filter (\(a, b) -> not (isMissing a) && not (isMissing b)) $ allAllelePairs of
+        [] -> Left $
+            "When checking for strand flips, I require at least one non-missing allele \
+            \pair to determine the consensus alleles. However, all allele pairs are missing for SNP " ++
```
nevrome: Does such a SNP stop the whole processing pipeline? I'm yet lacking the big picture.

It would be frustrating if these cases couldn't be removed automatically.

nevrome: I came across `joinEntryPipe` now. My understanding is that every `Left` here is reported, but does not interrupt the streaming. Good :+1:

test/testDat/testPackages/test_strandflips/POSEIDON.yml
```diff
@@ -0,0 +1,11 @@
+poseidonVersion: 2.4.0
```
nevrome: This is outside of the version range trident is supposed to read. Probably this would be an issue if this was used in the golden tests.

poseidon-hs.cabal
```diff
@@ -1,5 +1,5 @@
 name:                poseidon-hs
-version:             2.0.0.0
+version:             2.1.0.0
```
nevrome: Does this warrant a major version jump?

nevrome (commented):
I went through the code. Looks very good, overall. My main concerns are with the interface, but maybe I just don't understand exactly what you wanted to achieve.

src/Poseidon/Core/GenotypeData.hs
nevrome: You even write this out above:
> Any logging happens as a Left, which is then caught be the joinEntryPipe to report as error and resulting in skipping the SNP.

So nevermind.

src/Poseidon/CLI/Trident/OptparseApplicativeParsers.hs
stschiff: Yes, the setting right now is such that the default behaviour does _not_ change, and by default we i) assume that all SNPs live on the same strand, and ii) that we therefore do not have to check or flip anything.

There are arguments for setting a different default, I will raise this in a new comment below.

src/Poseidon/CLI/Xerxes/FStats.hs
stschiff: Hmm, yes, I expected this opinion from you 😁... and I was going back and forth on this. OK, let's discuss this below more fully.

src/Poseidon/Core/GenotypeData.hs
stschiff: Got it. Will change.

test/testDat/testPackages/test_strandflips/POSEIDON.yml
stschiff: Indeed. Fixed, and also found another test-pac with v2.4.0.

poseidon-hs.cabal
stschiff: Well, I felt it necessary because of the changed logging-messages. But I suppose that may not count as a breaking change, so probably better to make this a minor version change 👍 I will fix.

stschiff:
OK, so @nevrome raised three important points here:

## Should `--strandcheck` actually be the new default behaviour?

So, in [ADMIXTOOLS's `mergeit`](https://github.com/argriffing/eigensoft/blob/890ef8f24b6cf0b68e9dd11608f9a058a95ff2cd/CONVERTF/README#L256) it actually _is_ indeed the default to perform strandchecks. This means that by default, C/G and A/T SNPs are kicked out when merging two packages.

But is this a good default behaviour for trident? Hmm, well, it certainly is the safer option! But I think it would be a poor default, for the simple reason that our philosophy is to support package archives, with lots of packages that are constantly viewed together. It is reasonable to assume that the default case for genotype-merges are happening within well-defined multi-package archives, in which strand-checks can be excluded by construction.

In any case, making it the default would be a **_massive_** breaking change: Any command that involves merging, either inside forge or any other joint-loading, would now return a largely reduced dataset with A/T and C/G SNPs removed. I don't think we can do this.

## Should the strandcheck option be included in other commands that involve joint-genotype loading, such as inside xerxes? 

I voted against it, but maybe it should indeed be allowed. I am open for that

## If do _not_ make strandcheck the default, should there be some detection?

Yes, I think that is a great idea! Note, though, that we already warn here through the [`joinEntryPipe`](https://github.com/poseidon-framework/poseidon-hs/blob/bff9c9ddd6a9968c2ebfcf4dff74ca58ace181f7/src/Poseidon/Core/Package.hs#L740), which reports errors that emerged in the merging step. And [this line]() here actually produces an error if inconsistent SNPs are found. So we can consider two non-exclusive options:
1) Elevate this warning to an actual error, stopping the merging. This may be a problem if you're trying to merge a dataset that just has a few really faulty SNP entries that you would prefer just to be skipped. I don't know whether that's a real use-case. We could then also offer a `--skipErrors` option for people who know what they are doing.
2) We can add to the error message due to inconsistent alleles a line to possibly try `--strandcheck`, as a hint.

I think I like both of these option and propose to do them both now. I will test this with a real dataset that I know contains strand-flipped alleles. What do you think?

nevrome:
Ok - thanks for the explanation. I think what you already partially implemented in 46b17ac is the right way to go: Stop if incongruent SNPs are detected, and force the user to make a decision how to deal with them. That's the kind of hand-holding I appreciate.

If I understand correctly, then there may be issues that `--strandcheck` can not fix, right? So in certain situations one may want to call both `--strandcheck` and `--skip-incongruent-snps`? And this may then also apply for the xerxes subcommands?

I'm also wondering about the naming of these options:
1. `--skip-incongruent-snps` is fine and clear, but all other trident CLI arguments use camel case. So maybe  `--skipIncongruentSNPs` would be more consistent.
2. `--strandcheck` is misleading, because it does not only check, but also filter. When thinking about a better name I was even wondering if it does not just conflates two separate tasks?

stschiff:
Yes, agree to the renaming to camel case. Regarding the term `strandcheck`. That's the term used in admixtools, and I think I like it. It does three things:
- **check** whether alleles might be on different strands that need flipping.
- **fix** the strand if possible
- **filter** out sites that are strand-ambiguous. 

Those three tasks cannot be separated plausibly, so I don't think it "conflates" them. It really makes different assumptions and actions than in the normal case. I think `--strandcheck` is an OK word for this. Another possibility would be `--strandFlipMode` to indicate that it's a quite different merging action?

src/Poseidon/CLI/Xerxes/FStats.hs
stschiff: This is now resolved. I've added `--strandcheck` and `--skipIncongruentSNPs` to all subcommands that consume joint genotype data now.

stschiff:
OK, so I've completed this now. Things got more complicated (they always do), but also somewhat clearer and certainly safer. So now with default options, if people try to merge datasets with strand-flips, things should fail hard and throw an error about incongruent alleles, and a recommendation to either use `--skipIncongruentSNPs` or `--strandcheck` if they have a suspicion that strand flips might be the problem. I am testing this out now.

stschiff:
I've tested with real data and fixed some more edge cases. Ready for merge, from my end.

nevrome:
OK - nice! I think `--strandFlipMode` is far clearer, but considering the habits and familiarities of our users we should stick to `--strandCheck`.

To test the functionality at least once, I tried to forge the entire community archive into one package, first without `--strandCheck`. I got this error:

> Issues in genotype data parsing: PoseidonGenotypeException "Incongruent SNP entries for SNP: inconsistent alleles for SNP \\"rs6667923\\" at position 1:1206343. Consensus alleles inferred from all input packages are A, C, but package has alleles T, C. Could this be due to strand-flips? Consider using the --strandcheck option to check for strand flips and to automatically flip strands if needed.. Could this be due to strand-flips? If so, consider using --strandcheck. You can also consider using --skip-incongruent-snps to skip such SNPs."

The error message should probably be a bit less repetitive :wink:. That aside the question that immediatelly comes to mind is which packages cause this issue. I assume it would be a non-trivial modification to preserve this information for the error message, right?

With `--strandCheck` I got this error:

> Issues in genotype data parsing: PoseidonGenotypeException "Incongruent SNP entries for SNP: inconsistent alleles for SNP \\"rs6667923\\" at position 1:1206343. Consensus alleles inferred from all input packages are A, C, but package has alleles T, C. Could this be due to strand-flips? Consider using the --strandcheck option to check for strand flips and to automatically flip strands if needed.. Could this be due to strand-flips? If so, consider using --strandcheck. You can also consider using --skip-incongruent-snps to skip such SNPs."

It is regrettable that we have data in the community archive that can not be merged like this. It would be good to know what data that is... I also feel the error message should somehow reflect that I already tried `--strandCheck` (spelling mistake in the error message, btw.).

With `--skipIncongruentSNPs` the forge seems to work. At least the first 30k SNPs, then I aborted.

Shall I have a look myself into these things? You already invested a lot of time here, and I should probably just open a PR with my suggestions.

stschiff:
Yes, this is very useful feedback. Indeed we should improve the error message. I should also say that your feedback here inspires me to think of another matter:
Right now, we just check for inconsistencies across all packages that are to be forged. But of course, it is possible that 200 packages are fine, and only one has a third allele. In this case, I don't know whether we really want to exclude the entire SNP, or perhaps make somehow a majority judgment and flag only the data from the one offending package as missing. Or even more extreme: Only the individuals with the illegal third allele as missing? Hmm, hard to decide.

Perhaps as a start it would be indeed good which packages offend the merge and not make any more "smart" merging choices. Cause indeed, technically this really should not happen... I'm surprised that this happens with the community archive, actually, because I would have thought all of them use the same BIM file, more or less. 

Of course, your help would be welcome at the moment, as I am way behind many commitments, still, so I don't right now think I can get to this within the next days. Perhaps let's communicate via the Matrix chat in case one of us feels like implementing this further.

So the decision is:
- make the error reporting smarter, including package information, reduce redundancy in the message... 
- do _not_ make the merging more lenient for now.

nevrome:
Good - I can follow your argument. I agree that investigation is the first priority, so the error message should be improved. I also agree that the basic merging functionality should be strict and stop on anything abnormal.

Maybe eventually (probably not in this PR) we could think about strategies how to make the merging robust against individual outliers beyond `--skipIncongruentSNPs`. These could be available as additional CLI options.

nevrome:
Discussion in Poseidon meeting 2026-05-22: As the error message is thrown for iondividual SNPs and the streaming stops in this case, the error message can be very verbose. It should essentially list all SNPs for all packages to find a potential culprit.

nevrome:
This is good to merge and release. Just needs a little write-up for the release changelog. :+1:

stschiff:
OK, I've written the Release Text. Of course, that can still be edited after the merge. I hope I've described it fully. I will merge once the CI tests complete.

PR #381, opened 2026-04-30
Reactivate ploidychecks

stschiff:
This PR simply reactivates code that we had commented out, which checks whether the Ploidy-Information in the Janno agrees with the actual genotype data. If the Janno file claims that a particular sample should be `haploid`, but the genotype data contains heterozygotes, this will now throw an error. As far as I could test, all current packages in our archives pass this test, so I think it's safe to reintroduce. 

Note that this PR is on top of #375, so should be merged after that one.


nevrome:
I remember now why this feature was deactivated again in #275 after you first introduced it in #268. It's explained thoroughly in #276, but the gist of it is that we have old package versions in our archives that fail this check. They can not be served by the webserver any more if we activate it. I fear this is still unresolved. But at least we are now in a better position to implement a solution on top of #356 and #357.

stschiff:
I've implemented a version check for the ploidy checks, and also removed geno-checks from the server startup in general.

stschiff:
OK, I've now implemented a minimum PoseidonVersion 3.0.0 for ploidy checks. I've also removed genoChecks altogether from the server startup. Ready to review @nevrome

nevrome:
Looks good! I added a golden test to see it in action.

PR #383, opened 2026-05-18
added an extra executable just for the server

nevrome:
I decided to propose the creation of an own executable `poseidon-server` with special compiler settings to address grievances with our webserver VMs. With this `trident serve` is still available, but the server functionality also has its own dedicated space.


stschiff (approved):
Nice, thanks for doing this. Looks good to me.

PR #384, opened 2026-05-25
added a timelineplot to the html API

nevrome:
I have a version of this running on the internal server at <http://bioapp09.eva.mpg.de>.

After a number of experiments I decided to use the markercluster plugin for now, despite its obvious drawbacks. Rendering large amounts of data points individually is just not feasible with leaflet. The issue of inaccurate cluster point locations is mitigated by the fact that markercluster gives the polygon from which samples are aggregated when hovering a cluster. And clicking on a cluster conveniently zooms in on this relevant range.

Of course one could argue that leaflet is the wrong tool for such a plot in the first place, but it has two advantages: We don't need another dependency, and map and timeline plot have a very similar look and feel.

There are two more minor issues I should mention:
1. The timeline plot adds further loading time, which can already be an issue for very large archives.
2. On very low zoom levels the markerclusters sometimes seem to give inaccurate counts. I guess this is a bug, but the plugin is barely maintained, so there is not much hope for a quick fix.

Overall I think the plot is not ideal as it is, but certainly better than nothing. Have a look at the test setup and tell me if this is acceptable for you.


stschiff (approved):
This is great, and I have no comments. I obviously didn't check the Javascript in depth, but this is low-risk code, as it just affects the HTML. I have no problem approving this as is.

Thanks!

nevrome:
I'm not convinced any more that this particular plot is a good solution. Will think about it again. This PR does have some useful refactoring anyway, so I will continue here eventually.

PR #387, opened 2026-06-07
zip packages for download on demand

nevrome:
Not meant for merging into strandcheck, I just started from there and wanted to make the diff more readable.

I implemented the idea as discussed, which mostly just meant deleting code. It does work, but not in the lazy/streaming way we had hoped for. The data is now fully loaded in memory when a download is requested. More research is required to solve this.


nevrome:
I implemented streaming zipping now, using a library called [zip-stream](https://hackage.haskell.org/package/zip-stream). In my tests this worked remarkably well: Fast and with almost no memory-footprint. Could you please check if you can confirm this, @stschiff? Maybe you could also check if your macOS zipping tool can properly read the resulting archives.

stschiff:
Fantastic progress. Looking forward to testing this!

stschiff:
OK, it works impressively well. I've tried to monitor the memory footprint, but for Community archive packages it was so fast that I couldn't see much. I will go through the code now.

stschiff:
OK, went through the code. Must admit that some of the stuff looks a bit magic to me, but it all makes sense and it does seem to work. How do you suggest we deploy this? Shall we first fire up our test-server with this, and test a bit more thoroughly from there?

stschiff (approved):
I went through the code, and it makes sense to me. I don't understand 100% of it, though.

nevrome:
Setting up a test-server was a good idea. I discovered that the download progress indication in `trident fetch` does not work with this setup. Just as in the browser, trident does not know how big the file will be eventually.

nevrome:
Puh... further testing revealed more issues with `fetch`.

Some of the new AADR packages are very large. For large zip-archives >4GB a given zip implementation must support a feature called *zip64*. `zip-stream` automatically uses it when creating zip archive streams.

So far `fetch` continued to use `zip-archive` for unzipping of the downloaded archives, which does not support zip64. I therefore decided to implement the unzipping with `zip-stream` instead, and remove `zip-archive` entirely.

But that came with its own trouble: Unfortunately, `zip-stream`'s unzipping interface does not support a compression level of 0. So I had to turn on compression. Compression does not seem to increase the memory footprint on the server, but it is more CPU-intensive. Most annoyingly, the unzipping with `zip-stream` has a large memory-overhead on the client side: Fetching the AADR v66 2M dataset now requires 10GB of memory on the client, just to unzip.

All very unpleasant business, but I still think we should go ahead with this.

stschiff:
Wow, thanks for exploring! OK, so I don't quite understand why unzipping is now a problem. Can't we just revert to system calls in that case? I mean, we're downloading a file from the internet, and call unzip on it, isn't that it? Why do we even need Conduit or streaming-technology for this?

nevrome:
You mean we should just assume the user has some specific software tool installed for unzipping? That would indeed be much simpler. But trident is supposed to run on linux, macOS and windows. I don't see how we could do this elegantly...

I think the fundamental issue lies in the structure of .zip archives with critical information only at the end of the file (bad for streaming), and the lack of a zip library in Haskell that supports everything we need. Just to make this clear: The large packages are not supported by our previous solution `zip-archive`: it can't create proper archives for them. So we have to move to something else. We now learned that `zip-stream` also lacks one critical feature: unzipping without loading the entire archive into memory. There is yet another Haskell zip library we could try: https://hackage.haskell.org/package/zip. But that is a heavy dependency.

I wonder if we would fare better with tar. I know that would be a breaking change, but our server-client infrastructure has a way to make such changes safely. And for tar there seems to be a library that supports our usecase by Michael Snoyman himself: https://hackage.haskell.org/package/tar-conduit

nevrome:
Merged despite the memory issue. It's too important to have working fetch process, even if it's suboptimal.

PR #391, opened 2026-06-17
bumped sequence-formats dependency to allow for dots in bim files

stschiff:
This is a trivial PR that bumps the dependency on sequence-formats, which allows dots (`.`) to be read as allele characters.


nevrome:
Nice!

PR #393, opened 2026-07-09
Minor change to get proper output quoting in `writeJannoFileWithoutEmptyCols`

nevrome:
I noticed an issue in the output quoting of `writeJannoFileWithoutEmptyCols`. It caused the output of  `trident rectify --jannoRemoveEmpty` to be unreadable to trident! The minimal change in this PR does seem to solve the problem.


stschiff:
OK, I think I understand the change, looks OK to me. I'm a bit surprised about the back and forth here... why do we first encode the Janno into byte string, then decode again, filter for non-empty columns using transpose and lists, and then encode back again. Isn't that one loop too many? Why can't we just do the filter immediately on the Janno Rows? Is that due to the columns perhaps not being aligned and uniform? Probably, right?

stschiff (approved):

nevrome:
In their normal state the janno rows have all sorts of types and it's complicated to determine if they are NA or not. After encoding every NA value is a string `n/a`. So the easiest way to check if an entire column is NA is to encode, then decode to Bytestring. It's really just a hack, but it works and does the thing in a few lines of code.

PR #396, opened 2026-07-21
Various minor fixes/changes suggested by Thiseas and others

nevrome:
Changes collected here:
- Made the display of description and source url in the explorer page of the of the server's html API indendent of each other.
- Made the source column hide when the necessary URL is missing on the archive page.
- Added a input validation feature that checks line endings of the first (!) row in text files in packages. Only runs when checksums are present, because only then a difference in line endings can lead to unexpected behaviour.
- Made it possible to give ORCIDs with `--newContributors` in rectify.
- Fixed encoding issues in `trident list` when additional .janno columns are requested.
- Nicer error messages in case of missing input files in `validate`.


src/Poseidon/Core/Utils.hs
```diff
@@ -298,6 +303,34 @@ getChecksum f = do
     let md5Digest = md5 fileContent
     return $ show md5Digest
 
+-- helper function to check line endings of text files
+-- only considers the first line:
+-- if the byte immediately before the first LF is CR,
+-- then the first line uses CRLF; otherwise it uses plain Unix LF
+checkLineEnding :: FilePath -> PoseidonIO ()
```
stschiff, line 310: Wow this is a clever recursion.

test/PoseidonGoldenTests/GoldenTestsRunCommands.hs
```diff
@@ -457,6 +457,13 @@ testPipelineList testDir checkFilePath = do
         }
     runAndChecksumStdOut checkFilePath testDir (testLog $ runList listOpts7) "list" 7
 
+    let listOpts8 = listOpts1 {
+          _listRepoLocation  = RepoLocal [testPacsDirOther </> "Schmid_2028_utf8"]
+        , _listListEntity    = ListIndividuals (AddColList ["起司蛋糕"])
```
stschiff, line 462: delicious

test/testDat/testPackages/other_test_packages/Schmid_2028_utf8/Schmid_2028.janno
```diff
@@ -1,3 +1,3 @@
-Poseidon_ID	Group_Name	Genetic_Sex	Date_C14_Labnr
-XXX001	POP1	F	testà
-XXX002	POP2	F	MAMS-47224 
+Poseidon_ID	Group_Name	Genetic_Sex	Date_C14_Labnr	起司蛋糕
+XXX001	POP1	F	testà	鑽頭
```
stschiff, line 2: OK after the column title I thought you'd stick with the desert theme, but "drilling" and "sausage salad" aren't quite fitting. Anyway, good to have these tests!

stschiff (approved):
OK, went through this. Very nice! Thanks for fixing the nasty encoding bug in addCols. I had to drill a bit deeper to actually find it myself. I was very confused when List.hs had so few changes to it.

src/Poseidon/Core/Utils.hs
nevrome: I admit that  had some AI help for this one. Needed some back and forth to find a solution that reads as little as possible.

Issue #251, opened 2023-05-19
`forge` should get an option to append an optional .janno column with source-package information


nevrome:
This was raised a couple of months ago by @TCLamnidis and I thought I quickly formalize it into an issue. Let me know if there already is an issue about that in some of our repositories.

The idea is to add an option to forge to document forge operations on a sample/individual level in the .janno file, specifically with a new column that documents the source package the individual is extracted from.

That means a forge operation like this 

```
trident forge -d ./PacWithInd1 -d ./PacWithInd2 -f "<Ind1>,<Ind2>" -o outPac
```

should create a package `outPac` with a .janno file looking maybe something like this:

```
Poseidon_ID    ...    Source_Packages
Ind1           ...    PacWithInd1 -> outPac
Ind2           ...    PacWithInd2 -> outPac
```

The new column `Source_Packages` does not necessarily have to be included into the Poseidon schema -- we can discuss that. Also the name and the syntax are up for discussion.

If a package with this column is forged again, then we could simply append the next step in the column, so for example like this:

```
Poseidon_ID    ...    Source_Packages
Ind1           ...    PacWithInd1 -> outPac -> outPac2
Ind2           ...    PacWithInd2 -> outPac -> outPac2
```

Is this what you had in mind, Thiseas?

TCLamnidis:
roughly, yes. I was only looking for the original file, maybe with a version attached, rather than a complete reconstruction of the forge history of an individual (which I fear might look a bit messy after 3-4 forges). 
That said, this approach would more than cover my suggestion so 👍 from me conceptually.

stschiff:
Yes, that's great.

What I had in mind was mirroring Thiseas idea, which was just to record this the very first time, so that only the very original source packages are recorded. This could simply be implemented by forge checking if the column `Source_Packages` is already present in a row, and if so, do not alter it. 

The chain is a pretty cool idea. I am worried that it might become unwieldy though if people do many derived forges. Or is that just a too unlikely scenario?

Another disadvantage of the chain syntax is that we cannot simply use this column to filter on a given source package or so. 

How about we just drop the very last arrow? Then we would get after the first forge 

```
Poseidon_ID    ...    Source_Packages
Ind1           ...    PacWithInd1
Ind2           ...    PacWithInd2
```

which would then at least for the 90% of cases where there is only one forge make the output very easily understandable and filterable?

nevrome:
Yes - that sounds like a good compromise.

[...]

PR #397, opened 2026-08-10
implementing forge trace

stschiff:
This PR implements a Forge Trace, which addresses #251. 

The implementation is radically simple:
- _Every_ forge output which writes Janno, adds a new column "ForgeTrace" which lists for each sample the name and the package version that the sample came from.
- If `ForgeTrace` already exists, a new trace is appended with `;`.
- There is no special rule for one-package forges, so even subsetting packages results in a ForgeTrace. 

What do you think?
The package is not hard to review, the only relevant changes are in Forge.hs. But of course there are tons of Forge Test packages, which all had to be updated for the golden tests to pass.


stschiff:
I thought some more about it and decided to rename the new column to `SourcePackage` to make it a bit more tool-agnostic.

nevrome (commented):
Thanks for looking into this! @TCLamnidis will be happy to hear that this will be added soon. I just have three comments:

1. I think adding/updating this new column should not be the default, but a `forge` option. We could call it `--trace` or `--docSourcePac` or something like this. To illustrate why I would find this annoying as a default behaviour: I recently prepared the patched versions of the AADR v66 packages. There I wanted a minimal diff to keep things human-readable. A new column would have been very irritating in this context.
2. `SourcePackage` does not fit with the usual column naming scheme in Pascal_Snake_Case. Maybe we should go with `Source_Package`.
3. Maybe we should add a more complex test case where a package with the `Source_Package` column gets forged a second time, to see the concatenation in action. Maybe even from two different source packages.

stschiff:
Yes to all! One advantage of fixing 1) is that the golden tests won't all get updated. I'll get to it.

stschiff:
Done. Can you check again @nevrome ? Thanks

nevrome:
Great! I tested it and found it working :+1:.

There is a detail I think could be improved: When you merge a package A that already has a `Source_Package` column with a package B that does not, then the samples from B seem to get an extra `n/a` entry in the output package C. So these samples would be traced in C with `n/a;XXXX_Package-X.X.X` and not as expected with `XXXX_Package-X.X.X`. I'm not sure why this happens, but it is probably not the intended behaviour.

Beyond that I found the command line documentation a bit unclear. We also don't use `Janno` but `.janno file`, usually, but that is a matter of taste.

Finally, the feature should also be added to the release changelog text. When this is merged we can make a new trident release.

stschiff:
OK, agreed to all three points, will get them done. Thanks.

stschiff:
OK. Done. All three points addressed, added another Golden Test, and resolved the Branch conflict. Added Release Changelog text.

nevrome:
Excellent - thanks!

Issue #343, opened 2025-04-11
Rectify / Modify change


stschiff:
We will remake the current `rectify`. This whole process was started in a now closed PR #337 which ended with this final discussion point by @nevrome 

> Stephan and I had another discussion about this. Here's a quick summary of my understanding of the results:
> 
> The old, non-smart rectify was useful and should be preserved as is. It should get a new name (e.g. modify/alter/manipulate). Its flaw of editing every package below the base directory should be documented clearly. We could consider extra measures like a warning when editing multiple packages, or even a flag that has to be set in case multiple packages will be affected.
A new, smart subcommand rectify should be added. It automatically adds and corrects checksums and bumps the version of packages whose checksums have changed. It has only four options: -d (naturally), --packageVersion VPART and --logText STRING (mandatory), and --newContributors DSL (optional).

nevrome:
Looking at this again I think the new rectify should still allow to adjust checksums without incrementing the version number. Maybe we could make `--packageVersion VPART` and `--logText` optional (though if one is there, the other must be as well).

A good example where this is relevant is this commit here: https://github.com/poseidon-framework/community-archive/pull/281/commits/a72f6c57881f70f3312fac08e02527d94c3c80d6 The package author just wanted to change a technical detail of the .janno file in a draft package, which does not warrant a change in the version number. This common case should ideally be handled with a simple call to `trident rectify -d .`, right?

PR #401, opened 2026-09-03
split rectify and modify as decided in #343

nevrome:
As decided in #343 I renamed `rectify` to `modify` and then built a new `rectify` that simply searches for packages with changed checksums to automatically fix them. This is a breaking change, of course. This PR also includes new dependencies that speed up the MD5 sum computation considerably.

Remaining TODOs are:
1. ~~Add a warning or generally better command line output to `modify` to highlight that it changes all input packages.~~
2. ~~Add some golden tests for `rectify`.~~


stschiff (approved):
Nice, looks great, from my perspective. Just to be clear: the new rectify only detects changed checksums, not missing ones. That's intended, right?

